PR #82, opened 2019-08-08
minor style issues

Maarten-vd-Sande:


Maarten-vd-Sande:
@siebrenf I made a rule that every push to the master branch needs review. You are our honorary first reviewer of these minor style changes

Issue #73, opened 2019-08-07
in cluster environment every rule re-does the PE&SE lookup


Maarten-vd-Sande:

Maarten-vd-Sande:
https://stackoverflow.com/questions/57395931/snakemake-variable-that-defines-whether-process-is-submitted-cluster-job-or-the

PR #86, opened 2019-08-08
Store layout information in .snakemake. usefull for reruns, and potentially for #73.

siebrenf:
Also strips whitespace from sample names, for convenience.

### DIFF
--- a/rules/configuration.smk
+++ b/rules/configuration.smk
@@ -3,6 +3,8 @@ import re
 import time
 import subprocess
 import pandas as pd
+from ast import literal_eval
+from hashlib import blake2b
 from multiprocessing.pool import ThreadPool
 from snakemake.utils import validate
 from snakemake.logging import logger
@@ -11,6 +13,7 @@ from snakemake.logging import logger
 # read the samples file
 samples = pd.read_csv(config["samples"], sep='\t')
 validate(samples, schema=f"../schemas/{schema}")
+samples['sample'] = samples['sample'].str.strip()
 samples = samples.set_index('sample')
 samples.index = samples.index.map(str)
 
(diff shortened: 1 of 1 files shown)


rules/configuration.smk
```diff
@@ -41,39 +44,49 @@ assert sorted(config['fqext'])[0] == config['fqext1']
...
-    else:
-        results.append(tp.apply_async(get_layout, (sample,)))
-        # sleep 1.25 times the minimum required sleep time
-        time.sleep(1.25 / (config['ncbi_requests'] // 2))
-
-# now parse the output and store in config
-config['layout'] = {**config['layout'], **{r.get()[0]: r.get()[1] for r in results}}
+# layout file with a unique name depending on the samples contained
+identifier = str(blake2b(''.join(samples.sort_index().index).encode(), digest_size=12).hexdigest())
+layout_file = './.snakemake/layouts_' + identifier
```
Maarten-vd-Sande: So if I first try with samples:
GSM12345
GSM67890

And then again only with sample
GSM12345 

It needs to do the lookup again?

rules/configuration.smk
```diff
@@ -41,39 +44,49 @@ assert sorted(config['fqext'])[0] == config['fqext1']
...
+            layout[sample] = get_layout(sample)[1]
+            # sleep 1.25 times the minimum required sleep time
+            time.sleep(1.25 / (config['ncbi_requests'] // 2))
+
+    # save layout to file
+    with open(layout_file, 'w') as f:
+        f.write(str(layout))
+else:
+    with open(layout_file, 'r') as f:
+        layout = literal_eval(f.read())
```
Maarten-vd-Sande: not a fan of eval. Better would be to just pickle the dictionary

rules/configuration.smk
```diff
@@ -41,39 +44,49 @@ assert sorted(config['fqext'])[0] == config['fqext1']
...
+    layout = {}
+
+    # now do a request for each sample
+    for sample in samples.index:
+        if os.path.exists(expand(f'{{result_dir}}/{{fastq_dir}}/{sample}.{{fqsuffix}}.gz', **config)[0]):
+            layout[sample] = 'SINGLE'
+        elif all(os.path.exists(path) for path in expand(f'{{result_dir}}/{{fastq_dir}}/{sample}_{{fqext}}.{{fqsuffix}}.gz', **config)):
+            layout[sample] = 'PAIRED'
+        else:
+            layout[sample] = get_layout(sample)[1]
```
Maarten-vd-Sande: Now the lookups happen linear, the previous design allowed for parallel lookups

Maarten-vd-Sande:
Something to think about is where we want to store these files. If I first download the samples in the downloading pipeline, then align them in another, and then call peaks in a third, we do this lookup three times. Minor detail, but I'm open to suggestions

PR #105, opened 2019-08-21
get_genome can now download annotation.gtf and transcripts.fa

siebrenf:
(also removes a minor bug with genomepy plugins)

### DIFF
--- a/rules/configuration.smk
+++ b/rules/configuration.smk
@@ -119,6 +119,6 @@ def conda_path(yaml):
     with open(env_file, 'rb') as f:
         content = f.read()
     md5hash.update(content)
-    hash = md5hash.hexdigest()[:8]
-    path = os.path.join(env_dir, hash)
+    dir_hash = md5hash.hexdigest()[:8]
+    path = os.path.join(env_dir, dir_hash)
     return path
--- a/rules/configuration.smk
+++ b/rules/configuration.smk
@@ -104,3 +104,21 @@ if config.get('peak_caller', False) and 'hmmratac' in config['peak_caller']:
     assert all([config['layout'][sample] == 'PAIRED' for sample in samples.index]), \
     "HMMRATAC requires all samples to be paired end"
 
+
+# Find conda directories. Does not work with singularity.
+def conda_path(yaml):
+    """ Find the path to a conda directory """
+    import hashlib
+    import os.path
+
+    env_file = os.path.abspath(yaml)
+    env_dir = os.path.join(os.getcwd(), ".snakemake", "conda")
+
+    md5hash = hashlib.md5()
+    md5hash.update(env_dir.encode())
+    with open(env_file, 'rb') as f:
+        content = f.read()
+    md5hash.update(content)
+    hash = md5hash.hexdigest()[:8]
+    path = os.path.join(env_dir, hash)
+    return path
--- a/rules/get_genome.smk
+++ b/rules/get_genome.smk
@@ -1,5 +1,5 @@
 # the filetypes genomepy will download
-config['genome_types'] = ['fa', 'fa.fai', "fa.sizes"]
+config['genome_types'] = ['fa', 'fa.fai', "fa.sizes", "annotation.bed.gz", "annotation.gff.gz", "annotation.gtf"]
 
 rule get_genome:
     """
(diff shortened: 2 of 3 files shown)


rules/configuration.smk
```diff
@@ -104,3 +104,21 @@ if config.get('peak_caller', False) and 'hmmratac' in config['peak_caller']:
...
+
+    env_file = os.path.abspath(yaml)
+    env_dir = os.path.join(os.getcwd(), ".snakemake", "conda")
+
+    md5hash = hashlib.md5()
+    md5hash.update(env_dir.encode())
+    with open(env_file, 'rb') as f:
+        content = f.read()
+    md5hash.update(content)
+    hash = md5hash.hexdigest()[:8]
```
Maarten-vd-Sande: hash is a built-in Python function. This means you overwrite the function here (see difference in github colours). Better change the name

rules/get_genome.smk
```diff
@@ -17,17 +17,90 @@ rule get_genome:
         parallel_downloads=1
     priority: 1
     params:
-        config['genome_dir']
+        dir=config['genome_dir'],
+        gtf=expand("{genome_dir}/{{assembly}}/{{assembly}}.annotation.gtf.gz", **config),
```
Maarten-vd-Sande: Why is this a parameter, and not just output?

rules/get_genome.smk
```diff
@@ -17,17 +17,90 @@ rule get_genome:
...
     conda:
         "../envs/get_genome.yaml"
     shell:
         """
         active_plugins=$(genomepy config show | grep -Po '(?<=- ).*' | paste -s -d, -)
-        trap "genomepy plugin enable {{$active_plugins}} > {log} 2>&1" EXIT
+        trap "genomepy plugin enable {{$active_plugins,}} > {log} 2>&1" EXIT
```
Maarten-vd-Sande: Hooray :+1:

rules/get_genome.smk
Maarten-vd-Sande: Is required as input for rule get_transcripts. Has to be output even?

rules/get_genome.smk
```diff
@@ -17,17 +17,90 @@ rule get_genome:
...
+    output:
+        expand("{genome_dir}/{{assembly}}/{{assembly}}.transcripts.fa", **config)
+    log:
+        expand("{log_dir}/get_genome/{{assembly}}.transcripts.log", **config)
+    benchmark:
+        expand("{benchmark_dir}/get_genome/{{assembly}}.transcripts.benchmark.txt", **config)[0]
+    priority: 1
+    params:
+        path=conda_path("../../envs/get_genome.yaml"),
+        purgedfa=expand("{genome_dir}/{{assembly}}/{{assembly}}.purged.fa", **config)
```
Maarten-vd-Sande: Again why is this a parameter? It seems like it is output here. If it is 'nonsense' output you can always call temp() on the output and it will be deleted when not required anymore.

rules/get_genome.smk
```diff
@@ -17,17 +17,90 @@ rule get_genome:
...
+            for line in fa:
+                if line.startswith('>'):
+                    line = line.strip().split(' ')
+                    clean = True if len(line) == 1 else False
+                    break
+
+        if clean:
+            shell(conda_gffread + " -w {output} -g {input.fa} {input.gtf} >> {log} 2>&1")
+
+        # purge the dirty formatting from the fasta file, then use this in gffread
```
Maarten-vd-Sande: Generally comments are like this
```
if sth:
    # comment what to do
    ...
else:
    # another comment
    ...
```
or
```
# comment
if sth:
    ...
```

As opposed to
```
# comment what to do
if sth:
    ...
# another comment
else:
    ...
```

Maarten-vd-Sande: https://stackoverflow.com/questions/625708/where-to-put-comments-in-an-if-then-else-construct

rules/get_genome.smk
siebrenf: file is generated only when input fasta has dirty formatting

rules/get_genome.smk
siebrenf: will do!

PR #118, opened 2019-08-25
Cleanup of config.yaml.

Maarten-vd-Sande:
Before all configurations existed in the config. It can be confusing to see configuration logging with settings that are completely irrelvant. This PR should solve this. 

**What did change**
Changed the schemas to hold nested objects + added some parsing of these objects in configuration.smk + minor style issues.

Tests are updated accordingly

### DIFF
--- a/schemas/config/alignment_general.schema.yaml
+++ b/schemas/config/alignment_general.schema.yaml
@@ -32,25 +32,29 @@ properties:
   # aligner
   aligner:
     description: which aligner to use
-    default: bwa
-    enum: [bowtie2, bwa, hisat2, salmon]
-    type: string
+    type: object
+    properties:
+      bowtie2:  {type: object}
+      bwa:      {type: object}
+      hisat2:   {type: object}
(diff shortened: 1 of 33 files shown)


schemas/config/alignment_general.schema.yaml
```diff
@@ -32,25 +32,31 @@ properties:
...
+        align: ''
 
-  # salmon mapping
-  salmon_index:
-    description: optional arguments for salmon index, such as -k, --gencode and -decoys {decoyfile}
-    default: ''
-    type: string
-  salmon_aln:
-    description: optional arguments for salmon alignment
-    default: --seqBias --gcBias --validateMappings # --fldMean --fldSD (defaults are in use) # --posBias (currently experimental)
```
siebrenf: Defaults make the script way more user friendly. Otherwise time is wasted setting flasg, or they must be hardcoded...

Maarten-vd-Sande: See changes

Issue #120, opened 2019-08-27
FR: Implement stale bot


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
Those pesky stale issues! Lets get stale https://github.com/marketplace/stale

**Describe the solution you'd like**
A stale bot.

Maarten-vd-Sande:
https://github.com/probot/stale#is-closing-stale-issues-really-a-good-idea

PR #122, opened 2019-08-27
Create stale.yaml

Maarten-vd-Sande:
see #120

**What did change**
Added a stale.yaml

No, since this is completely on github side


siebrenf:
whatever floats your goat

Issue #98, opened 2019-08-19
Implement STAR aligner


siebrenf:
requires annotation.gtf, which should be accessible via Genomepy

siebrenf:
can also use an annotation file in .GFF3 format, as described on page 7 here: https://github.com/alexdobin/STAR/blob/master/doc/STARmanual.pdf

PR #124, opened 2019-08-27
Star aligner

siebrenf:
Implemented star aligner #98.

**What did change**
- star indexing
- star aligning
- get_genome: improved the generation of prerequisite files (annotation.gtf and transcripts.fa)

Yes. STAR aligning.

### DIFF
--- a/rules/alignment.smk
+++ b/rules/alignment.smk
@@ -214,7 +214,7 @@ elif config['aligner'] == 'star':
             gtf = expand("{genome_dir}/{{assembly}}/{{assembly}}.gtf", **config)
         output:
             dir = directory(expand("{genome_dir}/{{assembly}}/index/{aligner}", **config)),
-            tmp = temp(directory(expand("{genome_dir}/{{assembly}}/index/{aligner}_tmp", **config)))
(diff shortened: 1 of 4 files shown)


rules/alignment.smk
```diff
@@ -207,6 +207,72 @@ elif config['aligner'] == 'salmon':
...
+
+    rule star_quant:
+        input:
+            reads=get_reads,
+            index=expand("{genome_dir}/{{assembly}}/index/{aligner}", **config)
+        output:
+            dir=temp(directory(expand("{result_dir}/{aligner}/{{assembly}}-{{sample}}", **config))),
+            pipe=get_alignment_pipes()
+        log:
+            directory(expand("{log_dir}/{aligner}_align/{{assembly}}-{{sample}}", **config))
```
Maarten-vd-Sande: Should this be a directory?

siebrenf: Star creates a set of logs (5 files, each with different details), some of which are several MB large. 

By keeping all log files separate the experience is the same as for users of stand-alone star.
By storing all log files within the log/star_align directory, the file tree is not cluttered.

PR #126, opened 2019-08-28
Improved file tree clarity

siebrenf:
Fastqc files were placed in the same directory as the files they analyzed, resulting in unorganized directories. 

This PR adds a QC directory where the MultiQC output is placed, and places fastqc and trimming output in sub directories of the QC directory.

**What did change**
- The trimming rule now moves the trimming report output. 
- The fastqc rule now has a stricter input. 
- Merge_replicates has been given its own rule (it does not belong in get_fastq).
- Bug and stylefixes

No.


rules/trimming.smk
```diff
@@ -30,6 +30,12 @@ rule trim_galore_SE:
         if [ "{params.fqsuffix}" != "fq" ]; then
           mv "$(dirname {output.se})/{wildcards.sample}_trimmed.fq.gz" {output.se}
         fi 
+        
+        # move the trimming report to the desired directory
+        report=$(dirname {output.se})/{wildcards.sample}.{params.fqsuffix}.gz_trimming_report.txt
+        if [[ -f $report ]]; then
+            mv $report {output.qc}
```
Maarten-vd-Sande: If the report is not generated I want the rule to fail here, not that snakemake reports that the output is not generated (although this case should never happen.

rules/trimming.smk
```diff
@@ -62,7 +68,10 @@ rule trim_galore_PE:
         # now rename to proper output
         for f in $(find "$(dirname {output.r1})/" -name "{wildcards.sample}_*val_*.fq.gz"); do
             mv "$f" "$(echo "$f" | sed s/.fq/.{params.fqsuffix}/)"; done
-        for f in $(find "$(dirname {output.r1})/" -maxdepth 1 -name "{wildcards.sample}_*val_*.fastq.gz"); do
+        for f in $(find "$(dirname {output.r1})/" -maxdepth 1 -name "{wildcards.sample}_*val_*.{params.fqsuffix}.gz"); do
             mv "$f" "$(echo "$f" | sed s/_val_./_trimmed/)"; done
```
Maarten-vd-Sande: Oops :angel:

siebrenf:
> Looking good, I made a small change concerning moving the output.
> 
> What do you think about adding the other quality contol files to the qc folder? (samtools stats, featurecounts,mark_duplicates metrics).

That does make sense.

rules/trimming.smk
siebrenf: A similar system occurs elsewhere, I'll also remove it there then!

siebrenf:
I also tried to make default locations for directories (such as the fastq, qc and log directories) to be nestled in result_dir. 

If this would have worked, all instances of {results_dir}/{fastq_dir} could be reduced to {fastq_dir}. The benefit of this, is that the user could specify an output directory not nestled in {results_dir}, but the default directories are still nestled.

To do this one would have to iterate over all config files and schemas to check if the directories are the default, and then re-initiate them as nestled to result_dir (example `if config['fastq_dir] == 'default': config['fastq_dir] = config['result_dir] + '/' + config['fastq_dir']`). 

I did some tests with YAML anchors before which were not up to the task. Do you think this is worth pursuing @Maarten-vd-Sande?

Maarten-vd-Sande:
@siebrenf I think if you use a simple for loop it should be easy to implement. Something like this already exists:
https://github.com/vanheeringen-lab/snakemake-workflows/blob/5aa70605d0fede25be4f8dde26dab1bf4f034b0a/rules/configuration.smk#L54-L59

You can change it to simply append result_dir to each dir (except result_dir, log and benchmark, maybe im forgetting sth).

It could be that some wildcards or so do not work anymore, if so, then don't waste too much time on it. However it has the potential to make a lot of code much cleaner/shorter.

siebrenf:
> Have you tried if everything works? Travis seems to work, however did you actually run the pipeline to see if files get moved around properly?

Yes. There were some issues with trackhub, but these seem to be related to naming (UserWarning: shortLabel is limited to 17 characters in the browser, some characters will be truncated)

rules/configuration.smk
```diff
@@ -40,12 +40,14 @@ if 'condition' in samples:
                 f' condition {condition} and assembly {assembly}'
 
 
-# cut off trailing slashes and make absolute path
+# make absolute paths, cut off trailing slashes and nestle default directories in result_dir
```
Maarten-vd-Sande: Isn't it nest instead of nestled? :angel:

rules/configuration.smk
```diff
@@ -40,12 +40,14 @@ if 'condition' in samples:
...
 for key, value in config.items():
     if '_dir' in key:
         if key in ['result_dir', 'genome_dir', 'rule_dir']:
             value = os.path.abspath(value)
         config[key] = re.split("\/$", value)[0]
+        if key in ['sra_dir', 'fastq_dir', 'trimmed_dir', 'qc_dir', 'dedup_dir'] and key not in user_dirs:
+            config[key] = os.path.join(config['result_dir'], config[key])
```
Maarten-vd-Sande: So what happens if for instance sra_dir goes through the loop before result_dir? That will give us some bad parsing here I guess? I am pretty sure these loops should be split!

rules/configuration.smk
siebrenf: oops

rules/configuration.smk
siebrenf: you are right!

Issue #129, opened 2019-08-30
FR: parallel_downloads hardcoded resource


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
Confusing for new users, and practically NEVER more than one.

**Describe the solution you'd like**
Hardcode the resource (I don't know if we can though...)

PR #131, opened 2019-09-01
Set default resources (parallel_downloads & memory)

Maarten-vd-Sande:
See issue #129, and #7.
1. Basically users NEVER want mulitple downloads in parallel, so setting this value is inconvenient and doesn't add much.
2. Some rules (e.g. genrich) use a lot of ram, and the pipeline is bound by ram and not cpu. We still need to add memory usage to each rule (#7). I already implemented that when the user specifies no memory, the max of the computer is taken. 

**What did change**
I added a wrapper around the workflow.execute in configuration.smk. This is definitely undocumented and unadvised! However this is the only possible way, and seems robust. I tried changing the default_resources in the workflow, or global_resources, however did now work as intended. These attributes were also changed recently in the non-released version yet, so would immediatly be 'depricated'. 

Could have, no priority

### DIFF
--- a/rules/configuration.smk
+++ b/rules/configuration.smk
@@ -1,8 +1,11 @@
 import os
 import re
 import time
+import math
+import psutil
 import pickle
 import subprocess
+import numpy as np
 import pandas as pd
 from multiprocessing.pool import ThreadPool
 from snakemake.utils import validate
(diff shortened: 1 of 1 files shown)


Maarten-vd-Sande:
@siebrenf I pushed this through so I can switch branch to GitFlow #132

Issue #80, opened 2019-08-08
GSMs can contain multiple SRRs


siebrenf:
example: [GSM352203](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM352203)

Maarten-vd-Sande:
https://www.biostars.org/p/100325/

siebrenf:
Closing this issue as the last change seems to be the correct solution.

Maarten-vd-Sande:
When I am downloading GSM352203 it does not crash, however it seems to download only one of the 5 runs, and hence only dump this one.

siebrenf:
Are you sure the runs are not concatenated?

Maarten-vd-Sande:
Yes, the problem is with the wget fallback. Wget downloads all files each time for each id, and then seems to write them all to the same file

Maarten-vd-Sande:
Okay the current implementation is faulty, in that SRA's shouldnt be concatenated I believe. GSM3141725 consists of 2 runs, if I download that one through our pipeline I get a fastq of 2.2Gigabyte, however the two runs are 2.1 and 4.1 gigabyte.

We should concatenate them after the dumping I believe

PR #140, opened 2019-09-04
Fix i.a. #80

Maarten-vd-Sande:
When the wget fallback is used three things (can) go wrong related to files with multiple runs:
- mkdir fails since the dir already exists
- every run is downloaded for every run (so run ^ run downloads)
- downloaded runs overwrite eachother

**What did change**
- mkdir -p (does not fail when dir already exist)
- wget_url is search for 1 run, not 1 sample
- download to id name, not sample name


Maarten-vd-Sande:
How? Please make an issue about this, so this doesnt get lost.

siebrenf:
I do not know if this will really be the case. 

So far, only GSMs have had multiple SRRs within them. If ERR and DRR can have multiple runs within them, this would break on line 29 `IDS={wildcards.sample};`. However, as ERR and DRR should be [equivalent to SRR](https://www.biostars.org/p/248744/), I doubt it. 

I wrote the comment so the search function will take care of this if need be :)

PR #143, opened 2019-09-05
Jenkins

Maarten-vd-Sande:
Testing the pipeline in travis is kinda impossible. I setup a Jenkins server and now the tests are run on my computer! :tada: 

**What did change**
Mainly moving the tests from .travis.yaml to Jenkins/Jenkinsfile. Since I do NOT want multiple tests running at the same time on my computer I set in the jenkinsfile that only one can run per time. This has the disadvantage that the commit status only gets set to pending once your commit is running on the pipeline!
The initial tests are working but we need very **SMALL** genomes and samples. Currently I am working with the fugu assembly, which is small 300mb, but smaller is preferable. Best would be a model organism so we can have atac-seq, chip-seq, rna-seq etc. from the one assembly, but also the genome, and fastqs need to be small. 

I implemented tests for downloading and assembly of bwa, bowtie2, and hisat2. We need tests for a lot more, but this is a good start.


Maarten-vd-Sande:
Since I removed the .travis.yml this test ofcourse will never be run.. :angel:

PR #145, opened 2019-09-06
Jenkins

Maarten-vd-Sande:
- The README badge was from the jenkins branch, not master.
- Added a lockable resource to not have concurrent builds between branches
- This might(!) result in an immediate pending state, and build starting much later, which is good!


Maarten-vd-Sande:
502 BAD GATEWAY again from e-utilities.... Let's try again in a bit.

Issue #133, opened 2019-09-01
FR: Investigate feasability for version bot


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
We want to keep our environments updated for all rules, however checking this manually is a pain in the butt.

**Describe the solution you'd like**
See if an existing bot can be used. Otherwise could be fun/useful to implement one ourselves.

**Links**
https://github-bot-tutorial.readthedocs.io/en/latest/

PR #165, opened 2019-09-11
Conda bot

Maarten-vd-Sande:
See issue #133.

**What did change**
I added a conda-bot and let Jenkins periodically run it (every night).

Maybe.. We really need to make sure that we get all rules tested so nothing breaks accidentally.


Maarten-vd-Sande:
@siebrenf are the tests covering enough on your part to feel confident for auto updates?

siebrenf:
Yes!

Might I suggest to finish the blacklist PR, then pull develop into this branch and test the bot here?

Maarten-vd-Sande:
The bot already "works", in the sense that it's just a script that I will have to make a periodic Jenkins job. When it finds an outdated package it opens a PR and that will be tested. Here is an example of when I was working on it:
https://github.com/vanheeringen-lab/snakemake-workflows/pull/164

Maarten-vd-Sande:
And the blacklist PR is stuck until I find a solution to a weird snakemake edge case, or they fix it on their side.

Jenkins/conda-bot.py
```diff
@@ -0,0 +1,72 @@
...
+    pull_request_title = f'auto-update: {file.name}: '
+
+    lines = list(filter(None, file.decoded_content.decode("utf-8").split('\n')))
+    for line in lines:
+        # skip lines until the dependencies are listed
+        if not dependencies:
+            dependencies = 'dependencies:' in line
+        else:
+            # ignore lines starting with pip, deeper indented lines, and comments
+            if 'pip:' not in line and not line.startswith('    -') and not line.startswith('#'):
```
siebrenf: Ignore a wider range of text in env.yamls
```suggestion
            # ignore lines with pip, deeper indented lines, comments and empty lines
            if '  - ' in line and 'pip:' not in line and not line.startswith(('    -', '#', '\n')):
```

Maarten-vd-Sande:
@siebrenf I think perhaps I fixed the Jenkins issue with not properly doing tests when files changed :tada:

Issue #170, opened 2019-09-20
BUG: Jenkins does not report failure when DAG fails


Maarten-vd-Sande:
**Describe the bug**
http://ocimum.science.ru.nl/jenkins/blue/organizations/jenkins/Snakemake-Workflows/detail/develop/14/pipeline/22

PR #173, opened 2019-09-20
small stuff

Maarten-vd-Sande:
Cleaner snakefiles, cleaner config, Jenkins fails on failing tests...


Maarten-vd-Sande:
#170 fixed

Maarten-vd-Sande:
btw should also fix #169

PR #175, opened 2019-09-24
Rna seq

### LINKED ISSUE #171: BUG: [put title here]
**Describe the bug**
MultiQC sample name parsing doesn't handle assembly-sample names. Was inconsistent as well.

siebrenf:
fixed #171 
Also fixed several minor issues

**What did change**
added command line config to multiqc.

No


rules/assembly_hub.smk
```diff
@@ -63,5 +64,5 @@ rule bam_bigwig:
         deeptools_limit=1
     shell:
         """
-        bamCoverage --bam {input} --outFileName {output} --numberOfProcessors {threads} {params} --verbose >> {log} 2>&1
+        bamCoverage --bam {input.bam} --outFileName {output} --numberOfProcessors {threads} {params} --verbose >> {log} 2>&1
         """
```
Maarten-vd-Sande: .bai file not used if im correct?

workflows/atac_seq/Snakefile
```diff
@@ -1,16 +1,10 @@
...
-
 # complement the configuration
 configfile: "config.yaml"
 
 
-for schema in ['general', 'download', 'alignment_general', 'peakcalling']:
-    validate(config, schema=f"../../schemas/config/{schema}.schema.yaml")
-
 # do onstart/onexit things
+sample_schemas = ['sample', 'assembly', 'strandedness', 'condition']
```
Maarten-vd-Sande: I can't find where this is used

Maarten-vd-Sande: this = strandedness

rules/assembly_hub.smk
siebrenf: not in the command line, but the script fails if it is not present.

Maarten-vd-Sande:
Why is Jenkins here twice as check? branch and pr-merge?

Issue #181, opened 2019-10-01
FR: Support bedtools multicov for 'count tables'


Maarten-vd-Sande:
**Describe the solution you'd like**
Output a multicov bed/table with the difference between conditions

PR #185, opened 2019-10-11
implemented count table #181

Maarten-vd-Sande:
Make a simple count table for each peak, for further downstream analysis:
```
QNAME   start   end     GSM2837488      GSM2837489
FLLO01000001.1  100807  101247  153     64
FLLO01000001.1  433814  434214  336     864
FLLO01000001.1  987378  987867  572     114
FLLO01000001.1  987959  988938  2503    1400
FLLO01000001.1  1567413 1567833 116     669
```

**What did change**
Of all peaks a union is taken, and the number of counts under this union is taken for each sample.

Yes, not sure if it works with all conditions (e.g. merge, fisher, etc.)

### DIFF
--- a/envs/bedtools.yaml
+++ b/envs/bedtools.yaml
@@ -1,0 +1,7 @@
+name: bedtools
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - bioconda::bedtools=2.29.0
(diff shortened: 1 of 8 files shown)


Maarten-vd-Sande:
@siebrenf I merged the conflict

Maarten-vd-Sande:
I know, its a bit weird, but that's the standard naming convention.. I realized that most tools look for the bam index by .bam.bai, and not through .bai only.

Maarten-vd-Sande:
Btw the count table is quite dirty, in that it does not correct for sequence depth or anything. It is documented in the docs that it works this way.

Issue #186, opened 2019-10-16
FR: Separate rules for merging technical and biological replicates


siebrenf:
**Describe the solution you'd like**
Technical replicates can often be merged, while biological replicates need to be evaluated for batch effects. 

I think we want to give the user the option to name the column in samples.tsv, and reference that name in config.yaml. One option for both technical and biological replicates, and a secondary option of merging method.

Technical replicates could all be merged by concatenation (for now), and no default column name is needed. 
For biological replicates, the default column name could be 'condition' (although I think this is rather vague and would prefer 'replicates'). 
This way the user can merge technical and biological replicates by only this column in sampltes.tsv

**(optional) Additional context**
I'm currently changing the merge_replicate rule to all workflows.

siebrenf:
This is really needed for ATAC-seq workflows with both types of replicates.

siebrenf:
This is really needed for ATAC-seq workflows with both types of replicates.

siebrenf:
It's in!

PR #187, opened 2019-10-16
Merge replicates

siebrenf:
Replicate merging is now possible in all workflows.

If the user adds a column 'condition' in samples.tsv. samples will be merged based on this by default. Useful for technical replicates as described in #186.

**What did change**
All workflows now include merge_replicates.smk. qc.smk has been updated to always check for the merged samples if applicable.

Maybe. We could add a merging set of samples in every test.


rules/trimming.smk
```diff
@@ -19,8 +19,8 @@ rule trim_galore_SE:
     params:
         config=config['trim_galore'],
         fqsuffix=config['fqsuffix']
-    wildcard_constraints:
-        sample=f".*(?<!{config['fqext1']}|{config['fqext2']})+"
+    # wildcard_constraints:
+    #     sample=f".*(?<!{config['fqext1']}|{config['fqext2']})+" #superfluous since global wildcard constraint on wildcards.sample names
```
Maarten-vd-Sande: Just remove this :smile:

schemas/config/merge_replicates.schema.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+description: snakemake-workflows replciate merging configuration
+
+properties:
+  # replicate handling
+  combine_replicates:
+    description: which method to use to combine replicates (concatenation, idr or fisher)
+    default: merge # for technical replicates
+    enum: [fisher, idr, merge]
+    type: string
```
Maarten-vd-Sande: Why not add this to alighment_general schema?

siebrenf: can do, but then merge_replicates.smk should also be rolled into alignment.smk. Furthermore, this rule will be expanded upon when we allow for multiple types of merging, as per #186.

rules/trimming.smk
siebrenf: will do!

Issue #172, opened 2019-09-20
FR: Multi-genome assembly hubs


siebrenf:
daler/Trackhub can create a hub, does not create additional files.
Gaius-Augustus/MakeHub can create an assembly hub, but only for one genome.

Maarten-vd-Sande:
Are you working on this? Shall I take over?

Maarten-vd-Sande:
@siebrenf this can be closed right? It's working (good enough).

siebrenf:
I want to have a look at the last few things still. It may be a nice pre-christmass project :)

PR #188, opened 2019-11-04
Assembly hub support

Maarten-vd-Sande:
See #172 

WIP

**What did change**
WIP

WIP


Maarten-vd-Sande:
@siebrenf, what are the odds that the assembly is not supported by ucsc, but an annotation exists? Did you come across a situation like that?

Maarten-vd-Sande:
Something is wrong with the tests, it does not test everything :unamused:

Maarten-vd-Sande:
@Maarten-vd-Sande add mkdir to trackhub, reported that this sometimes causes issues?

rules/trackhub.smk
```diff
@@ -96,38 +99,169 @@ rule narrowpeak_bignarrowpeak:
...
+                                allow_redirects=True)
+        assert response.ok, "Make sure you are connected to the internet"
+
+        # see if the title of the page mentions our assembly
+        if not any(assembly == x for x in
+                   re.search(r'<TITLE>(.*?)</TITLE>', response.text).group(1).split(' ')):
+            trackfiles['twobits'].append(f"{config['genome_dir']}/{assembly}/{assembly}.2bit")
+
+        # now check if we have annotation for
+        # TODO
```
siebrenf: for later?

envs/ucsc.yaml
```diff
@@ -7,3 +7,4 @@ dependencies:
   - bioconda::ucsc-bedsort=377
   - bioconda::ucsc-bedGraphToBigWig=377
   - bioconda::ucsc-bedToBigBed=377
+  - bioconda::ucsc-fatotwobit=377
```
siebrenf: ucsc.yaml is for the track-/assembly hub? To me, the purpose is unclear from the name.

rules/trackhub.smk
Maarten-vd-Sande: Yes, its waiting for the annotation downloads and a bugfix from Snakemake (which they actually fixed last week :D). 

That part is necessary for nice assembly hub with genes, but not essential for assembly hub functionality

envs/ucsc.yaml
Maarten-vd-Sande: All the ucsc related tools are there, not the dependencies for the trackhub/assembly hub. Feel free to change it to sth more clear

Issue #193, opened 2019-11-05
FR: Multicov does not work for queryname sorted files (e.g. genrich)


Maarten-vd-Sande:
**Describe the bug**
We can not make a samtools index for queryname sorted files. This is bad, we should maybe always sort as well on coordinate?

PR #195, opened 2019-11-12
Added insert sizes to multiqc report

Maarten-vd-Sande:
A simple check for overdigestion is to take a look at insert sizes of paired-end reads. This is now implemented and outputed in the multiqc

**What did change**
If alignment happens, and the reads are paired-end, `picard CollectInsertSizeMetrics` is called on the bam file:
image

I had some troubles with R inside of picard, changing the channel priority seemed to fix it :smile:

### DIFF
--- a/envs/picard.yaml
+++ b/envs/picard.yaml
@@ -1,7 +1,7 @@
 name: picard
 channels:
-  - bioconda
   - conda-forge
+  - bioconda
   - defaults
 dependencies:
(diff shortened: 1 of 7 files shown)


Maarten-vd-Sande:
@siebrenf as we discussed today, this also implements that samtools-coordinate is ALWAYS(!) performed. This then also solves issue #193

Issue #191, opened 2019-11-04
FR: Blacklist


Maarten-vd-Sande:

Maarten-vd-Sande:
For issues with macs2
https://groups.google.com/forum/#!topic/deeptools/NiAhCRDjc9U

PR #196, opened 2019-11-14
WIP: JGASmits scATAC -> snakemake-workflows

Maarten-vd-Sande:
@JGASmits once you finished ''*your side*'' of the code you can merge it into the snakemake-workflows repository. It will be added to the (new) scATAC branch, so don't worry that you'll break anything.

Once merged I will apply the finishing touches and it will be added :tada: 

p.s. you can just keep working on your current branch, the pull request updates automatically


Maarten-vd-Sande:
Yes we definitely do not need the whole genomes :eyes: . 

Ideally we do this through genomepy:
https://github.com/vanheeringen-lab/snakemake-workflows/blob/master/rules/get_genome.smk#L12-L22

JGASmits:
How do you prune yaml files? I have no clue.
I merged the hisat2 env, it indeed worked with the one already present.

Python3 is a general env. I use to run some of the scripts and steps. I couldnt think of a mroe specific name. xD

Regarding the genome files I have removed them, you do need them in the downstream R analysis though.

This also makes genomepy implementation tricky, since you would also need to gather the correct ATAC blacklisted regions file, and promoter file for this downstream analysis. I think it would be possible but very tricky.

JGASmits:
Op een hg38 chromsize bestand na, deze is nodig om het SNAPobject aan te maken ;)

Maarten-vd-Sande:
@JGASmits Is there a difference between *regular* and atac blacklist? Adding blacklist regions is on the roadmap already: #191. I can add this one of these days. Where do you get this promotor file from?

Maarten-vd-Sande:
Okay cool. 

If you want I can do the `rule get_genome` part, since this will need to be reworked into a checkpoint, and I've been thinking already quite some about what approach to take into this.

JGASmits:
That would be nice, then ill start working on generating the proper files using the output of:

genomepy plugin enable blacklist
genomepy install hg38 UCSC --annotation

gr Jos

Maarten-vd-Sande:
btw I added [snaptools](https://bioconda.github.io/recipes/snaptools/README.html?highlight=snaptools#recipe-Recipe%20&#x27;snaptools&#x27;) to bioconda :smile:

Maarten-vd-Sande:
@JGASmits I accepted this one. And I will work on it a bit to do some smull stuff so its coherent with the other pipelines. If that's we'll release it into the master branch :smile:

PR #199, opened 2019-11-14
only log layout of current samples, fix #194

### LINKED ISSUE #194: BUG: configuration prints complete cache
**Describe the bug**
The cofiguration print at the start prints the complete cache, not only the samples in samples.tsv

Maarten-vd-Sande:
Solved #194


Maarten-vd-Sande:
Bad gateway, I am guessing its on their side :angel: . Ill retry in a bit

PR #201, opened 2019-11-18
Deseq fixes

siebrenf:
- bugfix in DESeq2 script
- fix for finding 0 differentially expressed genes crashing the DESeq2 script
- added warning to usage of underscores in DESeq contrast inputs
- bugfix in alignment Snakefile


rules/DE_analysis.smk
```diff
@@ -41,10 +41,13 @@ def get_contrasts():
...
                 new_contrast = batch + '+' + contrast[0] + '_' + reflvl + '_' + lvl if batch is not None else contrast[0] + '_' + reflvl + '_' + lvl
                 new_contrasts.append(new_contrast)
-        else:
+        elif l == 3:
             # remove '~', for uniformity
             new_contrast = original_contrast.replace("~", "").replace(" ", "")
             new_contrasts.append(new_contrast)
+        else:
+            print("Please do not use underscores in the samples file, column " + contrast[0] + '.\nDESeq2 suggests using only numbers, letters and periods.\n')
+            raise ImportError
```
Maarten-vd-Sande: I don't follow the logic how deseq and contrasts work tbh, but it's probably better to add general constraints in the configuration.smk (such as ./\$+ etc). 

I am not sure if you are looking in the samples.tsv or config.

`ImportError` -> `ValueError`

siebrenf: FYI: 
From samples.tsv it wants a column name and two or more unique rows from that column
These are specified in the config.yaml as column_group1_group2. Deseq2 throws a warning if other symbols are used, but will try to run the analysis anyway. 

I can move the check to the config, sure!

siebrenf:
A parser for the samples.tsv has been added.
-Leading and trailing whitespaces are stripped
-Other whitespaces are replaced with underscores 
-All characters except numbers, letters, underscores and periods are removed.
-Duplicate samples are marked

PR #202, opened 2019-11-19
Update configuration.smk

siebrenf:
added minus (-) to the list of accepted characters in samples.tsv

### DIFF
--- a/rules/configuration.smk
+++ b/rules/configuration.smk
@@ -30,7 +30,7 @@ assert sorted(config['fqext'])[0] == config['fqext1']
 samples = pd.read_csv(config["samples"], sep='\t')
 # sanitize column names
 samples.columns = samples.columns.str.strip().str.replace(' ', '_')
-samples.columns = samples.columns.str.replace('[^A-Za-z0-9_.]+', '', regex=True)
+samples.columns = samples.columns.str.replace('[^A-Za-z0-9_.\-]+', '', regex=True)
 for column in samples.columns:
     assert column[0:7] not in ["Unnamed", ''], \
         ("\nEncountered unnamed column in " + config["samples"] +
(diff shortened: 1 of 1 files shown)


rules/configuration.smk
```diff
@@ -30,15 +30,20 @@ assert sorted(config['fqext'])[0] == config['fqext1']
...
+    ("\n" + config["samples"] + " may only contain letters, numbers and " +
+    "underscores (_), periods (.), or minuses (-).\n")
+# samples.columns = samples.columns.str.replace('[^A-Za-z0-9_.\-]+', '', regex=True)
 for column in samples.columns:
     assert column[0:7] not in ["Unnamed", ''], \
         ("\nEncountered unnamed column in " + config["samples"] +
          ".\nColumn names: " + str(', '.join(samples.columns)) + '.\n')
 # sanitize table content
 samples = samples.applymap(lambda x: x.strip() if isinstance(x, str) else x)
 samples = samples.replace(' ', '_', regex=True)
```
Maarten-vd-Sande: Almost! Lets also not allow spaces in sample names!

rules/configuration.smk
```diff
@@ -30,15 +30,18 @@ assert sorted(config['fqext'])[0] == config['fqext1']
...
+assert not any(samples.columns.str.contains('[^A-Za-z0-9_.\-]+', regex=True)), \
+    ("\n" + config["samples"] + " may only contain letters, numbers and " +
+    "underscores (_), periods (.), or minuses (-).\n")
 for column in samples.columns:
     assert column[0:7] not in ["Unnamed", ''], \
         ("\nEncountered unnamed column in " + config["samples"] +
          ".\nColumn names: " + str(', '.join(samples.columns)) + '.\n')
 # sanitize table content
 samples = samples.applymap(lambda x: x.strip() if isinstance(x, str) else x)
-samples = samples.replace(' ', '_', regex=True)
```
Maarten-vd-Sande: I removed this one @siebrenf, or does that break anything?

Maarten-vd-Sande: Apparantly it does :angel:

Maarten-vd-Sande: The problem is that not everything is string in the samples dataframe, does it hurt to change everything to string? There is a column timepoint, do those have to be nrs or would string be okay?

siebrenf: That is exactly how it should happen! I actually added numbers in the example to make sure this exact thing would be interpreted right

PR #220, opened 2019-12-13
Update to master

Maarten-vd-Sande:
Let's not let changes pileup like last time + I would like to test the conda-bot.


Maarten-vd-Sande:
Somehow still skips tests?

siebrenf:
hell yea! Can't leave our 8 fans waiting :)

PR #232, opened 2019-12-17
Condabot fix

Maarten-vd-Sande:
Currently the conda-bot can update to versions outside of a channel, this should fix it.

The current builds are failing because r-base has a higher version in conda-forge.


Maarten-vd-Sande:
@siebrenf I forced this since the tests are not sensible for this anyways. And then I can now re-run just to check if it works.

Issue #228, opened 2019-12-16
BUG/FR: Data wrongly notated as single/paired-end


Maarten-vd-Sande:
See for instance:
- https://www.ncbi.nlm.nih.gov/sra?term=SRX5376909
- https://www.ncbi.nlm.nih.gov/sra/?term=GSM3607580
![image](https://user-images.githubusercontent.com/15323399/70889949-1abfbf80-1fe4-11ea-8bd0-c2ad72436ad1.png)

The pipeline very much does not like this. I emailed the author to update the layout, however to avoid future problems we could first try to parse the trace page, and if that doesn't work we use the supplied layout?
https://trace.ncbi.nlm.nih.gov/Traces/sra/?run=SRR8577639

siebrenf:
I see the issue ([how it should look](https://trace.ncbi.nlm.nih.gov/Traces/sra/?run=DRR002207)), but don't like the idea of having to parse every website... 

That said, the info is in a labelled column, which could be parse [as described here](https://www.simplified.guide/scrapy/scrape-table).

siebrenf:
wrong button

Maarten-vd-Sande:
By switching to pysradb we have this problem again. I am afraid we should just accept that we can not fix all problems, and that wrong database storage is not our responsibility.

PR #238, opened 2019-12-17
Badgeo

Maarten-vd-Sande:
As described in #228 

**What did change**
Checking for layout now happens both on the website as through the API. If the website is unambiguous then use website info, otherwise API. Did not seem to break anything so far on my part,


Maarten-vd-Sande:
p.s. GEO is complaining, ill rerun tests later

Maarten-vd-Sande:
For me the most sensible is to put the imports above the file it is used. Including duplicate imports. This is the way it is done in 'normal' Python, and I think we should try to do things as Pythonic as possible since everything is based on it.

PR #242, opened 2020-01-10
fix #241

### LINKED ISSUE #241: BUG: Bad SRRs
**Describe the bug**
SRR2778061-SRR2778064 give weird errors when trying to dump them.

Would be nice if supported :smile:

Maarten-vd-Sande:
Some SRRs gave weird errors when dumping. Problem seemed to be related to the sra being in the same folder as the fastq output (not entirely sure).

**What did change**
tmp output now not in same folder anymore.


Maarten-vd-Sande:
Now also shouldn't bother us anymore with the wrong line where the code crashes :tada:

PR #245, opened 2020-02-13
Gloryhammer

siebrenf:
**What's new?**
* Trackhubs now create additional information (when the assembly is not recognized by UCSC).
* Major bugfix in DESeq2 script where the reference level was assigned lexiconagraphically instead of by order in the design name.
* bam_to_bigwig now removed duplicate reads by default (can be changed in the config)
* Minor bugfixes & streamlining.


Maarten-vd-Sande:
Want me to take a look yet? Or prefer to play with Jenkins here?

Maarten-vd-Sande:
@siebrenf I am not a big fan of having too many tests in parallel on my computer. I don't want my work to be interrupted by tests :)

Maarten-vd-Sande:
The NCBI complaint is a problem of starting multiple workflows at the same time. Might need to be solved with a lock or sth.

siebrenf:
I've added my key. so you shouldn't be affected anymore. If this is stable, I'll leave it like this. If now the DAG test can go back to being linear.

btw, a large chunk of testing time is now going to installing miniconda and envs. 
Would it be possible to install it permanently, then start a script with conda update -q? 

The environments should update whenever the hash on the yamls changes, so if we can also install those locally, that would save a lot of time!

envs/snakemake-workflows.yaml
```diff
@@ -5,14 +5,15 @@ channels:
   - defaults
 dependencies:
   - pkgs/main::python=3.6
-  - bioconda::snakemake>=5.5.4
-  - conda-forge::pip=19.2.1
+  - bioconda::snakemake>=5.10
+  - pip=20.0.2
```
Maarten-vd-Sande: Not sure if the conda-bot works without channel, why did you remove it?

rules/configuration.smk
```diff
@@ -337,6 +337,10 @@ else:
         sample=any_given('sample', 'condition')
 
 
+def get_workflow():
+    return workflow.snakefile.split('/')[-2]
+
+
```
Maarten-vd-Sande: This is fine for now. I implemented sth like this as well in trackhub. I think however this will bite us in the ass later. Probably we should put it in some schemas or sth.

rules/trackhub.smk
```diff
@@ -183,13 +185,187 @@ rule twobit:
...
+    resources:
+        mem_gb=2
+    run:
+        with open(str(input.genome), "r") as genome_handle, open(str(output.mask_unsorted), "w+") as bed_handle:
+            p = Pool(threads)
+            # seqIO.parse returns contig data.
+            # Each contig is scanned by get_masked_regions (in parallel by imap_unordered).
+            # As soon as a contig is scanned, the output is yielded and written to file.
+            for softmasked_regions_per_contig in p.imap_unordered(get_masked_regions, SeqIO.parse(genome_handle, "fasta")):
+                bed_handle.write(softmasked_regions_per_contig)
```
Maarten-vd-Sande: Lol is this really the best way? That's how they do it in MakeHub?

envs/snakemake-workflows.yaml
siebrenf: when installing the env, git complains that you didn't add pip as a dependency. This fixes that complaint.

rules/trackhub.smk
siebrenf: yes, but without multiprocessing it. And I think they miss any masked regions at the end of the contig.

siebrenf:
Ill continue working on jenkins elsewhere. this PR is getting big enough xD

envs/snakemake-workflows.yaml
Maarten-vd-Sande: That's weird that this causes it... Seems like a bug on conda's side

PR #250, opened 2020-02-17
Develop

siebrenf:
bugfixes & optimizations galore!


siebrenf:
Jenkins now uses a unique NCBI key to prevent clashes

Maarten-vd-Sande:
Adding a mutex lock around layouts.p is I think a slightly better solution, but both work :).

siebrenf:
Yes, that would a future proof solution! I'll add it to the issues for later integration.

Maarten-vd-Sande:
Would be nice if there is some mutex lock that both bash and python understand. That way we could actually have multiple downloads at the same time (id2sra).

PR #252, opened 2020-02-20
Prettier badge

Maarten-vd-Sande:
Now that our testing results are so pretty, we should show them off to the world :wink: !


Maarten-vd-Sande:
Yikes, I branched from master, and did not expect all this mess to come along. Why/what is the difference between master and develop? Not sure which one is correct..

Maarten-vd-Sande:
I'll restart from develop, but keeping this open for now for the answer

siebrenf:
Master and develop should be identical. any changes (other than yours) are likely due to the... Uhm.. experiments with the revert feature.

Maarten-vd-Sande:
The differences are quite large

siebrenf:
All changes from #250 are already present, so it must be issues with Revert. Lets hope this will fix it!

PR #257, opened 2020-02-25
fix issue with short SRR ids

Maarten-vd-Sande:
I am having some issues with printf not working as desired. And this should fix that. 

`printf: 09,: value not completely converted`


Maarten-vd-Sande:
Let me first check if this also solves my issues with downloading on Cartesius

PR #259, opened 2020-02-27
Update trackhub.smk

Maarten-vd-Sande:
After our discussion of yesterday, should solve the issue of @JGASmits :tada:

### DIFF
--- a/rules/trackhub.smk
+++ b/rules/trackhub.smk
@@ -407,9 +407,13 @@ def get_trackhub_files(wildcards):
                                 allow_redirects=True)
         assert response.ok, "Make sure you are connected to the internet"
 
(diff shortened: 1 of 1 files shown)


siebrenf:
Clean solution. Nice!

(Can't figure out how to approve on mobile)

JGASmits:
This fixes the trackhub generation issue, but not yet the USCS figures out that it is GRCh38 right? :)

Maarten-vd-Sande:
No, the other way around. It should fix that patch numbers are being 'recognized' so that it does not influence our interaciton with the genome browser.

I believe @siebrenf fixed the trackhub issue in #258, it only needs to be merged.

PR #262, opened 2020-02-28
fix bug in get_genome

Maarten-vd-Sande:
The way we now parse genomepy config assumes there is at least one active plugin. Since gaps and sizes have been removed as plugins, this list is empty. This should fix that!


Maarten-vd-Sande:
oops I pushed to the wrong branch xD

Maarten-vd-Sande:
@siebrenf I forced this through since it isnt that interesting :)

PR #263, opened 2020-02-28
Cram support

Maarten-vd-Sande:
For mapping big (BIG!) datasets you do not want to store those as bam.

I added a config option named `cram_no_bam`, which makes a cram from your bam file. The documentation is already there:
https://github.com/vanheeringen-lab/snakemake-workflows/wiki/3.-Alignment#cram-support

In the future I will make every output temp :angel: and just keep what is in rule all. Thats why it is called cram_no_bam, since it shouldn't keep the bam as well.


Jenkins/alignment/cram_no_bam.yaml
```diff
@@ -0,0 +1,6 @@
+# tab-separated file of the samples
+samples: samples.tsv
+
+
+# how to sort bam
+cram_no_bram: true
```
siebrenf: typo: should be `cram_no_bam`

also, the comment could be clearer, maybe `# store aligned reads in cram format`?

Maarten-vd-Sande: Good catch.

PR #200, opened 2019-11-15
Alignmentsieve (blacklist, tn5 shift, minqual, rm mito reads) + prettier qc report

Maarten-vd-Sande:
**What problem is the PR solving / What's new?**
If genomepy manages to download a blacklist, it makes use of it.

**What did change**
WIP

PR #265, opened 2020-03-04
Develop into master

Maarten-vd-Sande:
I thought I start this PR already, since I want this to happen before we accept #200. 

What is happening here?! I might just force develop over master, since this is quite annoying.


Maarten-vd-Sande:
I merged master into develop and kept "ours" (develop). A quick scan didn't show anything weird.

Issue #264, opened 2020-02-28
FR:mutex lock around layouts.p


Maarten-vd-Sande:
If I want to start my snakemakes on the slurm cluster I do NOT want them to all read and write the layouts.p file at the same time :)

Maarten-vd-Sande:
Implement this
https://unix.stackexchange.com/questions/401415/mutex-and-semaphore-like-in-shell-script

PR #266, opened 2020-03-04
Mutex lock lookups

Maarten-vd-Sande:
Also see #264

### DIFF
--- a/rules/configuration.smk
+++ b/rules/configuration.smk
@@ -257,7 +257,7 @@ def get_layout_trace(sample):
 
 
 # try to load the layout cache, otherwise defaults to empty dictionary
-with Lock(layout_cachefile_lock):
+with FileLock(layout_cachefile_lock):
     try:
         layout_cache = pickle.load(open(layout_cachefile, "rb"))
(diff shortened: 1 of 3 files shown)


Maarten-vd-Sande:
This unfortunately does not solve the issue with parallel testing, since they refer to different locks. Slightly considering putting the lock at the root of the directory.

rules/configuration.smk
```diff
@@ -367,7 +371,8 @@ def convert_size(size_bytes, order=None):
 
 
 # by default only one download in parallel (workflow fails on multiple on a single node)
-workflow.global_resources.update({'parallel_downloads': 1, 'deeptools_limit': 1, 'R_scripts': 1})
+workflow.global_resources = {**{'parallel_downloads': 1, 'deeptools_limit': 1, 'R_scripts': 1},
+                             **workflow.global_resources}
```
Maarten-vd-Sande: Now we do not overwrite what the user supplies!

siebrenf: oh nice!

PR #267, opened 2020-03-05
Replicates

siebrenf:
Sample merging and combining of biological replicates in ATAC-peak calling must both be possible. This PR allows just that.

**What did change**
 - Sample merging can now be performed by `technical_replicates: merge`, which follows column `replicate` in samples.tsv. 
 - Biological replicates can be combined with `biological_replicates: idr/fisher`, which follows columns `condition` in samples.tsv.

**Additional changes**
- Simplified code logic in many places: instead of checking if replicates are merged everywhere and then either taking all samples in samples.tsv, or taking a subset each time, a new dataframe is made, `treps` (from technical replicates), to which can be referenced.
- Similarly, a dataframe with unique biological replicates, breps, is also made in `merge_replicates.smk`.
- A copy of samples.tsv and config.yaml is not made into the result_dir.

**Shortcommings**
Ideally, I would like to split merged samples per assembly. This way, all assemblies can have a (combined) sample named `stage12` for instance. The way fastqc is set up prevents this however, so each name in column `replicate` must be unique between assemblies.

In a perfect world, we would cycle through all the possible permutations of technical and biological replicates, using a sampleset containing both singular and multiple samples per technical and biological replicate. That would take hours however, so I'll add a dryrun version of this.

### DIFF
--- a/rules/call_peak.smk
+++ b/rules/call_peak.smk
@@ -1,11 +1,14 @@
 def get_genrich_replicates(wildcards):
-    assembly, sample_condition = "-".join(wildcards.fname.split('-')[:-1]), wildcards.fname.split('-')[-1]
-    if not 'condition' in samples.columns or config.get('combine_replicates', '') == 'merge' \
-    or (sample_condition in samples.index and not sample_condition in samples['condition'].values):
-        return expand(f"{{dedup_dir}}/{wildcards.fname}.sambamba-queryname.bam", **config)
+    """
+    keep: process individual files (default)
+    idr: process individual files, used by rule idr for combining (and moving to replicate_processed)
+    fisher: send list of files (can be length 1), then moved to replicate_processed by rule genrich_fisher
+    """
+    if 'condition' in samples and config['biological_replicates'] == 'fisher':  # != 'keep':
+        return expand([f"{{dedup_dir}}/{wildcards.assembly}-{replicate}.sambamba-queryname.bam"
+        for replicate in treps[(treps['assembly'] == assembly) & (treps['condition'] == wildcards.sample)].index], **config)
     else:
-        return expand([f"{{dedup_dir}}/{assembly}-{replicate}.sambamba-queryname.bam"
-        for replicate in samples[(samples['assembly'] == assembly) & (samples['condition'] == sample_condition)].index], **config)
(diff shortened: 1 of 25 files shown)


siebrenf:
fix #186

siebrenf:
Jenkins wont run, ill try to reopen the PR

rules/call_peak.smk
```diff
@@ -1,32 +1,26 @@
...
-        expand("{benchmark_dir}/genrich_pileup/{{fname}}.benchmark.txt", **config)[0]
+        expand("{benchmark_dir}/genrich_pileup/{{assembly}}-{{sample}}.benchmark.txt", **config)[0]
     conda:
         "../envs/genrich.yaml"
     params:
         config['peak_caller'].get('genrich', " ")  # TODO: move this to config.schema.yaml
-    threads: 15  # TODO: genrich uses lots of ram. Get the number from benchmark, instead of doing it through threads
+    threads: 15
+    resources:
+        mem_gb=7
```
Maarten-vd-Sande: It uses much, much more than 7. Maybe change it into 20 or so. Then you can also set 1 thread

rules/configuration.smk
```diff
@@ -75,11 +91,6 @@ samples.index = samples.index.map(str)
 if config.get('peak_caller', False):
     config['peak_caller'] = {k: v for k,v in config['peak_caller'].items()}
 
-    # if hmmratac peak caller, check if all samples are paired-end
-    if 'hmmratac' in config['peak_caller']:
-        assert all([config['layout'][sample] == 'PAIRED' for sample in samples.index]), \
-        "HMMRATAC requires all samples to be paired end"
-
```
Maarten-vd-Sande: Either we fully drop hmmratac, or we keep this. Not a bit of both

Maarten-vd-Sande:
p.s. I was planning to merge develop into master with a couple of small changes of the past time. I think this shouldn't go yet into master, since it's probable that there are quite some edge cases that fail. Same goes for #200

rules/configuration.smk
siebrenf: Turns out this section is moved, not deleted :)

Maarten-vd-Sande: Ooops :angel:

PR #269, opened 2020-03-10
Global mutex lock

Maarten-vd-Sande:
layouts.p is now stored in .config/snakemake.layouts.p. The filelock is now "global", in the sense that each user has a layouts and a lock around it. If people were to use ncbi keys on e.g. mb07, they should (?) never interfere with each other.

Since we have a global filelock we can re-use this for our id2sra rule, which means that the downloading can now we done in parallel!! (I set the default to 3)

Note that in the tests the downloading also happens in parallel :)

### DIFF
--- a/Jenkins/Jenkinsfile
+++ b/Jenkins/Jenkinsfile
@@ -67,26 +67,26 @@ pipeline {
                         # Do some dry runs to test the DAG
                         # test download workflow
                         printf "\ndownload default\n\n"
-                        snakemake -s workflows/download_fastq/Snakefile --directory workflows/download_fastq -n -j 48 --config ncbi_key=8d516bfc05a769280a23d8a2c0e08f94a708 --quiet
+                        snakemake -s workflows/download_fastq/Snakefile --directory workflows/download_fastq -n -j 48 --quiet
 
                         # test alignment workflow
                         # default config
                         printf "\nalignment default\n\n"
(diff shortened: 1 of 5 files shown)


Maarten-vd-Sande:
p.s. I have no clue why the commits from the previous PR came along. I tried switching branch but it didnt work. It doesn't matter since they are identical to develop anyways.

siebrenf:
Ok, so for my understanding:
- why don't we need the NCBI key anymore?
- the file lock prevents multiple layout searches, but what prevents multiple tests from downloading the same sample simultaneously?

Maarten-vd-Sande:
Nothing prevents downloading the same stuff at the same time except snakemake locking folders, but our previous solution didn't do anything about that either right?

Every time you do something that uses the e-utils (checking layouts and downloading sra), we now put a "lock" around the layouts.p file. When this lock exists other rules that want to use the e-utils will wait untill the lock is released.  We dont need the NCBI key anymore to bypass the problem of multiple searches.

Maarten-vd-Sande:
Wait, I just realized I am stupid. The downloads won't happen in parallel, they just block each other....

siebrenf:
wont the NCBI key speed up the rule that is running either way?

Maarten-vd-Sande:
It only speeds up the fetching of the layouts, which shouldn't take more than 10 seconds anyways.

Maarten-vd-Sande:
Its not pretty, but now it should work! This is me downloading two files at the same time :)
ezgif com-video-to-gif(2)

Issue #273, opened 2020-03-12
FR: Code linting


Maarten-vd-Sande:
WE NEED THIS
https://github.com/snakemake/snakemake/pull/276

siebrenf:
yes please!

PR #274, opened 2020-03-12
draft for linting

Maarten-vd-Sande:
We want linting!! Also #273


siebrenf:
It's live in snakemake 5.11.1!

Maarten-vd-Sande:
p.s. I added a line-length of 100. Not sure what is a sensible number, but I don't think we should ignore line lengths.

PR #276, opened 2020-03-13
Salmon

### LINKED ISSUE #212: update tximeta
Tximeta is at version 1.4.0 on conda. 

Until a newer version (>1.4.4) is added, quantification with salmon is not possible.

siebrenf:
- Re-implemented salmon support now Tximeta has been updated on conda (fix #212 ).
- All R scripts are now modified to handle the new replicate merging system.
- configuration.smk steps have been
    - improved: to accomodate the R scripts, the columns replicate and condition are removed from the dataframe samples unless they are actually used (which could be used to clean up a whole lot of `if x in samples and y != 'keep'` lines to just the first part...)
    - cleaned up: f-strings for readability, messages updated and newly added


siebrenf:
Second time I've had this. It has something to do with pushing while a test is already running. Reopening the PR should work...

PR #278, opened 2020-03-18
Small peak-calling fixes

Maarten-vd-Sande:
- Genrich back to two rules instead of four (sorry I couldn't resist)
- small fix with trackhub not using the correct peaks

### DIFF
--- a/rules/call_peak.smk
+++ b/rules/call_peak.smk
@@ -1,24 +1,33 @@
+def get_genrich_replicates(wildcards):
+    assembly_ish, sample_condition = "-".join(wildcards.fname.split('-')[:-1]), wildcards.fname.split('-')[-1]
+    assembly = assembly_ish.split("/")[-1]
+
+    if sample_condition in samples.index or (sample_condition in treps.index and config.get("technical_replicates", "") == "keep"):
+        print(sample_condition, "here")
+        return expand(f"{{dedup_dir}}/{wildcards.fname}.sambamba-queryname.bam", **config)
+    else:
+        return expand([f"{{dedup_dir}}/{assembly}-{replicate}.sambamba-queryname.bam"
+        for replicate in samples[(samples['assembly'] == assembly) & (samples['condition'] == sample_condition)].index], **config)
+
+
 rule genrich_pileup:
     """
-    Generate the pileup. For >1 sample, combined with Fisher's method, we use genrich_pileup_fisher 
-    
-    We do this separately from peak-calling since these two processes have a very different
+    Generate the pileup. We do this separately from peak-calling since these two processes have a very different
     computational footprint.
     """
     input:
-        expand("{dedup_dir}/{{assembly}}-{{sample}}.sambamba-queryname.bam", **config)
+        get_genrich_replicates
     output:
-        bedgraphish=expand("{result_dir}/genrich/{{assembly}}-{{sample}}.bdgish", **config),
-        log=expand("{result_dir}/genrich/{{assembly}}-{{sample}}.log", **config)
+        bedgraphish=expand("{result_dir}/genrich/{{fname}}.bdgish", **config),
+        log=expand("{result_dir}/genrich/{{fname}}.log", **config)
     log:
-        expand("{log_dir}/genrich_pileup/{{assembly}}-{{sample}}_pileup.log", **config)
+        expand("{log_dir}/genrich_pileup/{{fname}}_pileup.log", **config)
     benchmark:
-        expand("{benchmark_dir}/genrich_pileup/{{assembly}}-{{sample}}.benchmark.txt", **config)[0]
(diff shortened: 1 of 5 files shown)


rules/call_peak.smk
```diff
@@ -1,24 +1,33 @@
+def get_genrich_replicates(wildcards):
+    assembly_ish, sample_condition = "-".join(wildcards.fname.split('-')[:-1]), wildcards.fname.split('-')[-1]
+    assembly = assembly_ish.split("/")[-1]
+
+    if sample_condition in samples.index or (sample_condition in treps.index and config.get("technical_replicates", "") == "keep"):
+        print(sample_condition, "here")
+        return expand(f"{{dedup_dir}}/{wildcards.fname}.sambamba-queryname.bam", **config)
+    else:
+        return expand([f"{{dedup_dir}}/{assembly}-{replicate}.sambamba-queryname.bam"
+        for replicate in samples[(samples['assembly'] == assembly) & (samples['condition'] == sample_condition)].index], **config)
```
siebrenf: should this not be
`for replicate in treps[(treps['assembly'] == assembly) & (treps['condition'] == sample_condition)].index], **config)`
to get the replicate names (if they exist)?

Also, this whole section assumes `condition` is in samples.

rules/call_peak.smk
```diff
@@ -1,24 +1,33 @@
+def get_genrich_replicates(wildcards):
+    assembly_ish, sample_condition = "-".join(wildcards.fname.split('-')[:-1]), wildcards.fname.split('-')[-1]
+    assembly = assembly_ish.split("/")[-1]
+
+    if sample_condition in samples.index or (sample_condition in treps.index and config.get("technical_replicates", "") == "keep"):
+        print(sample_condition, "here")
```
siebrenf: remove

rules/trackhub.smk
```diff
@@ -611,42 +611,41 @@ rule trackhub:
...
-                                )
-                                priority += 1
-                                trackdb.add_tracks(track)
-
-                        for sample in treps[treps['assembly'] == assembly].index:
-                            bigwig = f"{config['result_dir']}/{peak_caller}/{assembly}-{sample}.bw"
-                            assert os.path.exists(bigwig), bigwig + " not found!"
-                            sample_name = f"{sample}{peak_caller}BW"
+                        for brep in breps[breps['assembly'] == assembly].index:
+                            bigpeak = f"{config['result_dir']}/{peak_caller}/{assembly}-{brep}.bigNarrowPeak"
```
siebrenf: what if you called your biological replicate the same as your technical replicate/sample?

this section of code can accept both, which may lead to ambiguity. the only way to get that ambiguity out (I think), is separating the outputs to different folders.

rules/call_peak.smk
```diff
@@ -1,24 +1,33 @@
+def get_genrich_replicates(wildcards):
+    assembly_ish, sample_condition = "-".join(wildcards.fname.split('-')[:-1]), wildcards.fname.split('-')[-1]
+    assembly = assembly_ish.split("/")[-1]
+
+    if sample_condition in samples.index or (sample_condition in treps.index and config.get("technical_replicates", "") == "keep"):
```
siebrenf: can be simplified to 
`if sample_condition in treps.index:`

rules/call_peak.smk
```diff
@@ -1,24 +1,33 @@
+def get_genrich_replicates(wildcards):
```
siebrenf: this function becomes ambigous if a biological replicate has the same name as a technical replicate/sample. this is why i split the rule outputs by folder.

If I'm mistaken, or there is a solution, it would be really nice!

Maarten-vd-Sande: But you really shouldn't have the same names for those things right? I could add an assertion in configuration. It also becomes unclear in the trackhub which peak/bigwig belongs to what, and in the multiqc report.

rules/call_peak.smk
Maarten-vd-Sande: ooops :angel:

rules/call_peak.smk
Maarten-vd-Sande: Yep

rules/call_peak.smk
Maarten-vd-Sande: Or unless you have a single sample for a condition?

siebrenf: that would work. I left it as an option, but this would indeed remove possible confusion... go for it!

rules/call_peak.smk
Maarten-vd-Sande: I think I actually should use `treps_from_brep` here.

rules/trackhub.smk
Maarten-vd-Sande: The only thing I changed here is to make sure that the peak is plotted next the the bigwig track it belongs, not first all peaks, and then all bigwigs

PR #279, opened 2020-03-19
Better defaults

Maarten-vd-Sande:
Simple as that. @siebrenf I don't know much about rna-seq, should we add anything here?


siebrenf:
yep, ill take a look at it!

If we're going to expose useful config options in our yamls (which I agree is a smart thing to do), I think it might be very helpful to add `# fastq_dir: (default is inside result_dir)` to all of them.

Maarten-vd-Sande:
Yep

siebrenf:
have a look and see if you can approve my format changes. i tried to make things standard between workflows

siebrenf:
I'm getting the same error locally. The weird thing is, I still get it with 

```
rule setup_blacklist:
    output:
        expand("{genome_dir}/{{assembly}}/{{assembly}}.customblacklist.bed", **config)
    shell:
        'touch {output}'

rule alignmentsieve:
    input:
        bam=rules.samtools_presort.output,
        bai=expand("{result_dir}/{aligner}/{{assembly}}-{{sample}}.samtools-coordinate-unsieved.bam.bai", **config),
        blacklist=rules.setup_blacklist.output
```
The reference `rules.setup_blacklist.output` clearly works. It just doesn't want to run `setup_blacklist`?

The test also worked for Bowtie? I originally thought that may have to do with the temp(), but removing it didn't help here.

Maarten-vd-Sande:
It seemed to work for me locally. The weird thing is this worked in the original pull request, but after merge into develop not anymore. I'll take a look tomorrow, since this seems a blacklist thingy

Maarten-vd-Sande:
D'ohh I didnt see that it works for 1 out of 4 aligners... wth

Maarten-vd-Sande:
--notemp seems to fix the issue, now something happens in rna-seq and I have no clue :angel:

Maarten-vd-Sande:
By the way it is kinda confusing for me that in the samples.tsv the columns are condition and replicate but in the config biological and technical replicates. I feel like we should use one term (no preference for either).

Maarten-vd-Sande:
So the minimal mapq values (min_mapping_quality) values are really annoying, because aligners are not adhering to the same standard. I removed the value for star because you said something about that deseq does it already? 30 for bwa means no multimapping, and I think that corresponds to 255 for star.

siebrenf:
You are correct. 255 for no multimappers in STAR. 

This value could be tweaked with `--outSAMmapqUnique 30`, or we could set the value to 255 for the RNA-seq config since STAR is the only aligner available to it. I remembered reading about STAR removing multimappers, but I think that was for STAR solo, so ignore that.

Do we want to be this stringent by default though? I'm all for having this option in the config, but maybe commented out? Maybe something like:

`# min_mapping_quality: 255 (only keep uniquely mapped reads from STAR alignments)`

Maarten-vd-Sande:
Idk, I just asked Simon what he does for bwa, and he said 30, so no multimappers. Not sure if that is common in rna-seq though. I have no clue what the standards are

Maarten-vd-Sande:
Shall we merge this?

PR #283, opened 2020-03-23
marked lots of files as temp

Maarten-vd-Sande:
Now we store much much less intermediate files, + multiqc report is optional just like trackhub.

One thing to discuss is, do we want fastq files to be stored? I think not.


siebrenf:
Is a file marked as temp, if the rule that produces it isn't run? Most of our colleagues have local fastqs that they would not like to lose, so to safeguard that the answer depends on this question.

If not, then sra2fastq your output temp() files.
If yes, then we could add a flag to delete fastqs, and do this as a final step in trimming (if flag is true).

Maarten-vd-Sande:
Okay this works much worse than I expected!! It indeed removes files it didn't generate itself, and even worse, in this example you end up without any output since it is all deleted!

```
rule all:
    input:
        [f"file_{i}.txt" for i in range(5)]


rule tmp:
    output:
        temp("file_{sample}.txt")
    shell:
         "touch {output}"
```

I'll remove the temp flag around files that is output of any of the workflows, since otherwise you end up with nothing...

PR #287, opened 2020-03-24
Cpulimit

Maarten-vd-Sande:
On cartesius cpulimit doesn't work... Now you can call snakemake like this:

`snakemake --profile ocimum --config cpulimit=False`

I will add a small comment somewhere in the docs


Maarten-vd-Sande:
Yeah at first I also thought I'll just secretly add this. But it is nice for when we want tor run it on cartesius, now it is semi-documented :stuck_out_tongue:

PR #290, opened 2020-03-26
Qc stuff

Maarten-vd-Sande:
Added some macs2 quality values to multiqc and made it prettier :)

+ some stuff that was atac specific now wont pop up in chip

### DIFF
--- a/rules/qc.smk
+++ b/rules/qc.smk
@@ -6,7 +6,7 @@ rule samtools_stats:
     Get general stats from bam files like percentage mapped.
     """
     input:
-        expand("{dedup_dir}/{{assembly}}-{{sample}}.{{sorter}}-{{sorting}}.bam", **config)
+        expand("{result_dir}/{aligner}/{{assembly}}-{{sample}}.samtools-coordinate-unsieved.bam", **config)
     output:
         expand("{qc_dir}/dedup/{{assembly}}-{{sample}}.{{sorter}}-{{sorting}}.samtools_stats.txt", **config)
     log:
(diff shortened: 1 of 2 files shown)


rules/qc.smk
```diff
@@ -6,7 +6,7 @@ rule samtools_stats:
     Get general stats from bam files like percentage mapped.
     """
     input:
-        expand("{dedup_dir}/{{assembly}}-{{sample}}.{{sorter}}-{{sorting}}.bam", **config)
+        expand("{result_dir}/{aligner}/{{assembly}}-{{sample}}.samtools-coordinate-unsieved.bam", **config)
     output:
         expand("{qc_dir}/dedup/{{assembly}}-{{sample}}.{{sorter}}-{{sorting}}.samtools_stats.txt", **config)
```
siebrenf: if the input is no longer in dedup, neither should the output be

rules/qc.smk
```diff
@@ -253,14 +253,24 @@ def get_alignment_qc(sample):
     output.append(f"{{qc_dir}}/dedup/{{{{assembly}}}}-{sample}.samtools-coordinate.metrics.txt")
     output.append(f"{{qc_dir}}/dedup/{{{{assembly}}}}-{sample}.samtools-coordinate.samtools_stats.txt")
 
-    # add insert size metrics
-    if config['layout'][sample] == "PAIRED":
-        output.append(f"{{qc_dir}}/InsertSizeMetrics/{{{{assembly}}}}-{sample}.tsv")
+    if "atac_seq" in get_workflow():
```
siebrenf: why not for the other workflows anymore? MT DNA ratio seems a useful QC mark for all of them to me, or is it to cumbersome?

Maarten-vd-Sande: I haven't seen seen anyone mention this for RNA or CHiP. For ATAC it is important since it cuts open DNA. And mitochondrial DNA is very open, and there are many mitochondria. You want to make sure you didnt only pcr your mtochondrial reads. We could keep it, but I prefer the multiqc reports to contain pipeline specific outputs.

Maarten-vd-Sande: Let's just keep it for now, but ideally the multiqc reflects the pipeline that was run.

siebrenf: sure. this was a question, not a request :)

PR #291, opened 2020-03-26
Broadpeaks

Maarten-vd-Sande:
Now if you add `--broad` to a macs2 call it actually works :tada:


JGASmits:
I'm afraid the last commit: "please work" did not listen to its title. The broad peaks runs perfectly on my data. However in the trackhub generation rule: 'peak_bigpeak' it crashes with all the files because:

pass1 - making usageList (45 chroms): 3 millis
Error line 4786 of /home/jsmits/Epithelial_fate_KC_LSC/Chipseq_broad/macs2/GRCh38.p13-Chipseq_LSC_TERT_H3K27me3.tmp.broadPeak: score (1282) must be between 0 and 1000

Which I think  is the reason it crashes.
Perhaps we could write a small (awk) script to replace all the vallues larger than 1000 with 1000?

I'm gonna take a look at how the called braod peaks look like next!

JGASmits:
I've (manually since the trackhub generation didnt work :( ) generated a genome browser session from it to vizualize some broad peaks on H3K27me3 and H3K27ac data:
https://genome-euro.ucsc.edu/s/jossmits_3%40hotmail.com/broadpeaks

I think generally the called broad peaks look nice! I think the broad peak locations give a good general impression on where the histone marks are present. However I dont have much experience with broad peak calling (yet), so I'm curious what @simonvh thinks about the peak calling?

simonvh:
Looks kind of OK for K27ac, but the H3K27me3 peaks are not optimal. It currently calls smaller peaks within broad domains. Ideally it should call the whole domain as a peak.

Maarten-vd-Sande:
Okay awesome, I agree with Simon that the peaks do not look so good though :angel:... These regions should be a single broad peak.. Apparantly at least for HOMER it is quite sensitive to the parameters:
image
So I hope that that's the case for macs2 as well :)

PR #294, opened 2020-03-28
Control

Maarten-vd-Sande:
You can now add a column control and fill it with a name of a fastq, or sample from online. This will then be added as a control :)

**What did change**
One thing to note is that empty values in this column are allowed, which makes for some stupid checks sometimes for nans/strings. But for the rest quite straightforward.


rules/call_peak.smk
```diff
@@ -33,7 +44,7 @@ rule genrich_pileup:
     computational footprint.
     """
     input:
-        get_genrich_replicates
+        unpack(get_genrich_replicates)
```
siebrenf, line 47: in line 64 you call the input `{input.treps}`, so maybe call this `treps = unpack(get_genrich_replicates)` ?

siebrenf: nvm, its a dict. im stupid

rules/get_fastq.smk
```diff
@@ -165,6 +165,7 @@ def get_wrong_fqext(wildcards):
     # exclude samples with the correct fqext
     r1 = wildcards.sample + "_" + config["fqext1"] + "." + config["fqsuffix"] + ".gz"
     r2 = wildcards.sample + "_" + config["fqext2"] + "." + config["fqsuffix"] + ".gz"
+    print(wildcards)
```
siebrenf, line 168: remove

Maarten-vd-Sande:
I started first with control and let @JGASmits test that for me :angel:. And then I started working on the qc here, and decided to already incorporate the control as well. So it got a bit messy.

Maarten-vd-Sande:
Everthing here should be present in the other one as well

PR #296, opened 2020-03-30
Deeptools qc

Maarten-vd-Sande:
Added some extra quality control output to the multiqc:
image

**What did change**
Just added some rules that add the deeptools qc. This already depends on the control pull request, so that should be merged before this one.


Maarten-vd-Sande:
This has problems again with checkpoints :cry: ... UGH

Maarten-vd-Sande:
@siebrenf I changed a bit the parallel_deeptools thingy. Not sure if more threads on a single process works better than spreading them out. Do you have an opinion on that? It doesn't really matter on my pc, but apparantly on mb07 it barely does anything with too much deeptools.

rules/configuration.smk
```diff
@@ -417,11 +424,12 @@ def any_given(*args):
         elif column_name is 'sample':
             elements.extend(samples.index)
 
+    elements = [element for element in elements if isinstance(element, str)]
```
siebrenf, line 427: samples is loaded with dtype = 'str', so this shouldn't be needed (?)
can be left in tho

Maarten-vd-Sande: So the PR's got a bit mixed up, but I added the option to add a control to your samples.tsv. This field can also be left empty, so that's why I check if it's a string, since nan isnt.

Maarten-vd-Sande: sample assembly control
GSM123 hg38 GSM321
GSM432 hg38

rules/configuration.smk
```diff
@@ -459,7 +472,7 @@ def convert_size(size_bytes, order=None):
 
 
 # by default only one download in parallel (workflow fails on multiple on a single node)
-workflow.global_resources = {**{'parallel_downloads': 3, 'deeptools_limit': 1, 'R_scripts': 1},
+workflow.global_resources = {**{'parallel_downloads': 3, 'deeptools_limit': 16, 'R_scripts': 1},
```
siebrenf, line 475: I have no idea if this is still the case, but rule bam2bigwig (which uses deeptools BamCoverage) was only running 1 process at the time, no matter how or what. If that is no longer the case, then I'm totally fine with upping the limit.

Maarten-vd-Sande: `deeptools_limit` is now the max number of parallel deeptools threads. So yes :).

siebrenf: (in reference to your question:) this is also why I was allocating so many threads to one deeptools job

Maarten-vd-Sande: On my own computer I don't have much issues with it, but on mb07 it indeed gets very slow with more threads. My hope/guess is on cn45 and cn46 this won't be much of an issue because of newer/reformatted disks...

siebrenf: That's not what I meant. sorry.

I meant that without a deeptools limit, only one bamCoverage would run at once. SO I set the limit so that the threads would not be reserved for tasks that cannot run in parallel.

If that is not happening for bamCoverage anymore, then go ahead and keep this change.

Maarten-vd-Sande: So what I think happens is that not only 1 deeptools runs, but that it somehow has a very very big impact on IO, and that the computer can not write/read at that pace. And that's why it seems like nothing is happening. The way I now set it is that deeptools_limit is the max number of deeptools threads that can run in parallel. For instance 4 rules * 4 threads each, or 1 rule 16 threads. BamCoverage now uses 16 threads, so this should work.

Maarten-vd-Sande: image

Here is an example of multiple alignmentsieves doing almost nothing on mb07 because of the IO. I can check on my own computer and indeed confirm that alignmentsieve is waiting a lot for IO (`iotop`), however i can not check and confirm on mb07 because no root access

siebrenf: Interesting solution! I look forward to seeing it in action :)

PR #300, opened 2020-03-31
update genomepy

siebrenf:
to fix ensembl genomes


siebrenf:
working on a big refactor now. its why my activity here has dropped :)

PR #310, opened 2020-04-14
Small patches

Maarten-vd-Sande:
small patches (two bug fixes, including the fastqc fix) + alignmentsieve removed and just do the sieving ourselves with samtools and custom tn5 shift script.

Seems to do the same as alignmentsieve, a small difference in how the cigar string is parsed. Both alignmentsieve and this script do it wrong, since the CIGAR string is nonsensical when shifting reads. Only edge-cases should be different. Alignmentsieve also removes reads which the mate hasn't been mapped. The differences between implementations are very very minor:
http://ocimum.science.ru.nl/redo_danrer/multiqc_danRer11_sieve.html#deeptools_correlation  # own sieving
http://ocimum.science.ru.nl/redo_danrer/multiqc_danRer11.html#deeptools_correlation  # alignmentsieve

A quick scan in the genome browser doesn't show differences where reads start/end:
image

Also I don't notice differences between eyeing the sams (one by alignmentsieve and one by our script)
image


Maarten-vd-Sande:
The sieving took 13 minutes for a 6 gigabyte bam file. So that should be acceptable speed :), but no IO bottleneck :tada:

Maarten-vd-Sande:
I halving the threads :)
image

rules/configuration.smk
```diff
@@ -109,6 +109,8 @@ for schema in sample_schemas:
 samples = samples.set_index('sample')
 samples.index = samples.index.map(str)
 
+print(samples)
```
siebrenf, line 112: rm

rules/qc.smk
```diff
@@ -448,7 +448,7 @@ def get_peak_calling_qc(sample):
     # deeptools profile
     assembly = treps.loc[sample, "assembly"]
     # TODO: replace with genomepy checkpoint in the future
-    if assembly.lower() in ["ce10", "dm3", "hg38", "hg19", "mm9", "mm10", "grch38.p13", "grch38"]:
+    if assembly.lower() in ["ce10", "dm3", "hg38", "hg19", "mm9", "mm10", "grch38.p13", "grch38", "danrer11"]:
```
siebrenf, line 451: what's going on here? It looks like the blacklist genomes, but there's Ensembl builds, and zebrafish?

Maarten-vd-Sande: Checking if an assembly has an annotation. We can't do this with checkpoints unfortunately.. The list is definitely incomplete. Maybe genomepy has an api?

siebrenf: There will be a unified method of obtaining the annotation URL (if it exists) in the next version. Currently there is none. Get Simon to accept the PR :p

That said, I get your intentions and this would indeed be a good point for a checkpoint.

rules/qc.smk
```diff
@@ -162,7 +162,9 @@ def get_descriptive_names(wildcards, input):
 
     labels = ""
     for file in input:
-        trep = re.findall(f"{wildcards.assembly}-([^\.]+)\.+?", file.split("/")[-1])[0]
```
siebrenf, line 165: this truncated a name with a period in there. New code truncates after the `.sam` to capture both `.samtools` and `.sambamba`

PR #324, opened 2020-05-03
Only chromosome name + annotation

Maarten-vd-Sande:
GRCh38's annotation and chromosome names do not overlap, so deeptools qc tool complains.


siebrenf:
yep, only those UCSC genomes have a blacklist :(

Issue #319, opened 2020-05-01
BUG: IDR doesn't work


Maarten-vd-Sande:
**Describe the bug**
result_dir is recognized as a wildcard

![image](https://user-images.githubusercontent.com/15323399/80804081-63070500-8bb4-11ea-939a-12176782fd0d.png)

PR #329, opened 2020-05-07
Idr fix

Maarten-vd-Sande:
see #319


Maarten-vd-Sande:
I didn't start from branch develop, so now it seems more is changed than actually is. Only the changes in `call_peak.smk` are relevant. And they are quite irrelevant :angel:

PR #336, opened 2020-05-26
seq2science migration

Maarten-vd-Sande:
In the seq2science branch we collect all the changes for our migration from the name of snakemake-workflows to seq2science.

Once these are done "*all*" we have to do is merge, rename the repository and check all urls.


Maarten-vd-Sande:
Closed for now to not show non-migration commits and files.

PR #360, opened 2020-06-03
continuous deployment

Maarten-vd-Sande:
All we have to do is tag a commit with a (semantic) version, and it will become a release on github. Which later can be automatically picked up by bioconda :tada:


Maarten-vd-Sande:
After the "release" of seq2science I think it might be good to also keep a changelog.

Maarten-vd-Sande:
Yes I agree> I'm merging it, we can just add stuff to the changelog after we released

PR #369, opened 2020-06-04
Fix#355

### LINKED ISSUE #355: BUG: RNA-seq - decoy aware salmon indexing does not work

siebrenf:
fix #355


siebrenf:
Also fixes tests with incorrect paths (aligners and some trackhub tests)

PR #371, opened 2020-06-04
Multiqc thingies

Maarten-vd-Sande:
We now add the samples and config to the multiqc report :tada: 
ezgif com-video-to-gif


siebrenf:
question: should we add the `config.yaml`? Currently, when I run my script I use a profile, and config and maybe one or two command line flags. 

Shouldn't we instead convert the `config` dict back into a yaml, and add that? (Same question goes for the logger btw)

Maarten-vd-Sande:
It's really really ugly though

Maarten-vd-Sande:
+ those changes are also not stored when you copy the configfile to the log dir.

PR #373, opened 2020-06-05
use seq2science instead of snakemake

Maarten-vd-Sande:


siebrenf:
s2s now also accepts lists as --snakemakeOption

PR #381, opened 2020-06-08
Samstats

Maarten-vd-Sande:
Samtools stats before and after quality sieving. This means that the effect of certain steps can be traced, instead of getting good/bad results without knowing what effect sieving had.

**What did change**
Samtools stats gets called twice, once right after alignment and once sieving. The mutliqc schema now is cookiecutter-like, in that it accepts variables from the environment.


Maarten-vd-Sande:
Wait for (hopefully) a solution: https://github.com/samtools/samtools/issues/1266

PR #384, opened 2020-06-09
Update docs.yml

Maarten-vd-Sande:


siebrenf:
I like what you're doing, but this is altering the actual code. So shouldn't this be merged into develop?

PR #387, opened 2020-06-11
optional mamba support

Maarten-vd-Sande:
- Maybe we should print that we recommend to install mamba when it's not installed?
- Should we have a cool banner? :scream:

### DIFF
--- a/bin/seq2science
+++ b/bin/seq2science
@@ -20,6 +20,12 @@ try:
 except ImportError:
     full_import = False
 
+try:
+    import mamba
+    conda_frontend = "mamba"
+except ImportError:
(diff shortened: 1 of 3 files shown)


siebrenf:
The try/except is not required. Snakemake seems to switch to conda silently. That said, it could be used to print the suggestion to use mamba, which I agree would be nice.

You could even put a subprocess `conda install mamba -c conda-forge` there... but that may be a bit much...

Maarten-vd-Sande:
I don't think you should ever install something in someone's environment like that :angel: 

I will add a print, and maybe a cool banner ;)

Maarten-vd-Sande:
image

PR #388, opened 2020-06-11
Correlation_plot as png

Maarten-vd-Sande:
Currently WIP. 

Can't collapse different contents: http://ocimum.science.ru.nl/multiqc/multiqc_report.html


Maarten-vd-Sande:
This is still relevant. I am hoping/waiting for a new release of multiqc where I can collapse custom content and modules together. Phil Ewels (the maintainer) thought this was already possible.

This would be nice so I can put the correlation plots together in the deeptools section.

Maarten-vd-Sande:
Now replaces the not-clustered interactive heatmaps with clustered non-interactive images

image

Maarten-vd-Sande:
p.s. I gave up waiting for dendrogram support in multiqc :upside_down_face:

Maarten-vd-Sande:
p.p.s. I should do the colors the other way around!

Maarten-vd-Sande:
@siebrenf this is ready for review

siebrenf (approved):
Nice! 
Maybe the RNA-seq correlation plot can go with this in, but I'll compare the two later.

Maarten-vd-Sande:
What do you mean? The order of the image? I can do that.

This is based on pileup over the genome. Not sure how useful, it just gives a general overview. Once differential peaks has been implemented, we should add these plots for the peaks as well, since that is a much more obvious way of calculating it.

PR #397, opened 2020-06-12
Small changes

### LINKED ISSUE #394: BUG: annotation file
**Describe the bug**

```
        # unzip annotation if downloaded, warn if required but empty
        if [ -f /mnt/c/Users/simon/Documents/prj/test_seq2science/genomes/bosTau9/bosTau9.annotation.gtf.gz ]; then
            gunzip /mnt/c/Users/simon/Documents/prj/test_seq2science/genomes/bosTau9/bosTau9.annotation.gtf.gz >> /mnt/c/Users/simon/Documents/prj/test_seq2science/results/log/get_genome/bosTau9.genome.log 2>&1
        elif echo /mnt/c/Users/simon/Documents/prj/test_seq2science/genomes/bosTau9/bosTau9.fa /mnt/c/Users/simon/Documents/prj/test_seq2science/genomes/bosTau9/bosTau9.fa.fai /mnt/c/Users/simon/Documents/prj/test_seq2science/genomes/bosTau9/bosTau9.fa.sizes /mnt/c/Users/simon/Documents/prj/test_seq2science/genomes/bosTau9/bosTau9.gaps.bed | grep -q annotation.gtf; then
            echo '
Annotation for bosTau9 contains no genes. Select a different assembly or provide an annotation file manually.

' > /mnt/c/Users/simon/Documents/prj/test_seq2science/results/log/get_genome/bosTau9.genome.log
            exit 1
        fi

        (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)
```

The pipeline fails as there is no annotation file. However, there is:

```
 $ ls  /mnt/c/Users/simon/Documents/prj/test_seq2science/genomes/bosTau9/
README.txt  bosTau9.annotation.bed.gz  bosTau9.annotation.gtf
```

And the `atac_seq` pipeline should be able to run without gene annotation?

simonvh:
* Updated docs.
* Fix #394

### DIFF
--- a/seq2science/rules/get_genome.smk
+++ b/seq2science/rules/get_genome.smk
@@ -55,7 +55,7 @@ rule get_genome:
         if [ -f {params.gtf}.gz ]; then
             gunzip {params.gtf} >> {log} 2>&1
         fi
-        if [ ! -f {params.gtf} && echo {output} | grep -q annotation.gtf ]; then
+        if [ ! -f {params.gtf} ] && [ echo {output} | grep -q annotation.gtf ]; then
             echo '\nAnnotation for {wildcards.assembly} contains no genes. Select a different assembly or provide an annotation file manually.\n\n' > {log}
             exit 1
         fi
--- a/seq2science/rules/get_genome.smk
+++ b/seq2science/rules/get_genome.smk
@@ -42,7 +42,7 @@ rule get_genome:
         genomepy plugin disable {{blacklist,bowtie2,bwa,star,gmap,hisat2,minimap2}} >> {log} 2>&1
         genomepy plugin enable {{blacklist,}} >> {log} 2>&1
         
-        # download the genome and attempt to download the annotation
+        # download the genome and attempt to download the annotation and blacklist
         if [[ ! {params.provider} = None  ]]; then
             genomepy install --genomes_dir {params.dir} {wildcards.assembly} {params.provider} --annotation >> {log} 2>&1
         else
@@ -51,13 +51,20 @@ rule get_genome:
             genomepy install --genomes_dir {params.dir} {wildcards.assembly} NCBI    --annotation >> {log} 2>&1
         fi
         
-        # unzip annotation if downloaded and gzipped, warn if required but empty
+        # unzip annotation if downloaded and gzipped
         if [ -f {params.gtf}.gz ]; then
-            gunzip {params.gtf} >> {log} 2>&1
+            gunzip -f {params.gtf}.gz >> {log} 2>&1
         fi
-        if [ ! -f {params.gtf} ] && [ echo {output} | grep -q annotation.gtf ]; then
-            echo '\nAnnotation for {wildcards.assembly} contains no genes. Select a different assembly or provide an annotation file manually.\n\n' > {log}
-            exit 1
+        
+        # if assembly has no annotation, or annotation has no genes, throw an warning
+        if [ ! -f {params.gtf} ] || $(grep -q "No genes found" {log}); then
+            echo '\nEmpty or no annotation found for {wildcards.assembly}.\n' | tee -a {log}
+            
+            # if an annotation is required, make it an error and exit.
+            if $(echo {output} | grep -q annotation.gtf); then
+                echo'\nSelect a different assembly or provide an annotation file manually.\n\n' | tee -a {log}
+                exit 1
+            fi
         fi
         """
 
--- a/seq2science/rules/get_genome.smk
+++ b/seq2science/rules/get_genome.smk
@@ -30,7 +30,7 @@ rule get_genome:
     params:
         dir=config['genome_dir'],
         provider=config.get('provider', None),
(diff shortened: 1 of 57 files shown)


Maarten-vd-Sande:
Not sure why some other previous changes of me came along...

btw I feel like I have been chaning back and forth between zipped and non-zipped blacklists and annotations. Has this been changed on the genomepy side? 

About ATAC-seq without annotation. That is fully correct! I don't know why this has changed, I will take a look.

seq2science/rules/get_genome.smk
```diff
@@ -55,7 +55,7 @@ rule get_genome:
         if [ -f {params.gtf}.gz ]; then
             gunzip {params.gtf} >> {log} 2>&1
         fi
-        if [ ! -f {params.gtf} && echo {output} | grep -q annotation.gtf ]; then
+        if [ ! -f {params.gtf} ] && [ echo {output} | grep -q annotation.gtf ]; then
```
simonvh, line 58: @Maarten-vd-Sande I don't understand this test.

If I'm correct, it checks if `annotation.gtf` is in the output files. If it is, then it gives the warning that the annotation contains no genes. What is the purpose here?

Maarten-vd-Sande: @siebrenf

simonvh:
> Not sure why some other previous changes of me came along...
I merged the `develop` branch in my branch. Maybe I shouldn't have done that.

>btw I feel like I have been chaning back and forth between zipped and non-zipped blacklists and annotations. Has this been changed on the genomepy side?
Yes, it has. However, I think the current solution should always work. It now always expects an unzipped gtf. If it is zipped by genomepy it will unzip it. So this now shouldn't depend on behavior of genomepy.

simonvh:
> About ATAC-seq without annotation. That is fully correct! I don't know why this has changed, I will take a look.

However! An additional quality control could be enrichment of ATAC-seq reads in promoters. Does depend on quality of the annotation though.

Maarten-vd-Sande:
>Yes, it has. However, I think the current solution should always work. It now always expects an unzipped gtf. If it is zipped by genomepy it will unzip it. So this now shouldn't depend on behavior of genomepy.

Yes I noticed. Elegant solution :+1: 

> However! An additional quality control could be enrichment of ATAC-seq reads in promoters. Does depend on quality of the annotation though.

One way we use the annotation for atac-seq is to plot the profile with deeptools, which comes near-ish to what you describe (https://deeptools.readthedocs.io/en/develop/content/tools/plotProfile.html). One issue with snakemake however is that checkpoints are not working for us unfortunately (some weird edge-case bug), so currently I just hardcoded a list of some assemblies for which the annotation is downloaded:
https://github.com/vanheeringen-lab/seq2science/blob/master/seq2science/rules/qc.smk#L550

simonvh:
>One issue with snakemake however is that checkpoints are not working for us unfortunately (some weird edge-case bug), so currently I just hardcoded a list of some assemblies for which the annotation is downloaded:
> https://github.com/vanheeringen-lab/seq2science/blob/master/seq2science/rules/qc.smk#L550

I think there is always annotation for Ensembl and UCSC genomes. So maybe that knowledge helps.

seq2science/rules/get_genome.smk
siebrenf: In older versions of genomepy, an annotation could be downloaded without any content. genomepy then gave a message something like "... annotation.gtf contains no genes!"

this is no longer the case, so i need to change this a bit

siebrenf: oh right. it also throws an error if the gtf is in the output files, but not found. Since only some workflows have it in the output files! New checks are almost ready

seq2science/rules/get_genome.smk
```diff
@@ -51,12 +51,20 @@ rule get_genome:
...
-            echo '\nAnnotation for {wildcards.assembly} contains no genes. Select a different assembly or provide an annotation file manually.\n\n' > {log}
-            exit 1
+        # unzip annotation if downloaded and gzipped
+        if [ -f {params.gtf}.gz ]; then
+            gunzip -f {params.gtf}.gz >> {log} 2>&1
+        fi
+        
+        # if assembly has no annotation, or annotation has no genes, throw an warning
+        if [ ! -f {params.gtf} ] || $(grep -q "No genes found" {log}); then
+            echo '\nEmpty or no annotation found for {wildcards.assembly}.\n' | tee -a {log}
```
Maarten-vd-Sande, line 61: @siebrenf why not redirect to log? 
`>> {log}`

siebrenf: Because if you get this error at the start of your pipeline, its not too late to pick a different provider. So I think it might be nice to actually get the error, as well as getting it in the log.

Maarten-vd-Sande: ok :+1:

Maarten-vd-Sande:
@simonvh & @siebrenf shall we merge this?

Issue #391, opened 2020-06-11
BUG: I don't have 12 cores :-(


simonvh:
**Describe the bug**

```
 seq2science run atac_seq --cores 4
Traceback (most recent call last):
  File "/home/simon/anaconda3/envs/seq2science/bin/seq2science", line 283, in <module>
    main()
  File "/home/simon/anaconda3/envs/seq2science/bin/seq2science", line 47, in main
    _run(args, base_dir, workflows_dir, config_path)
  File "/home/simon/anaconda3/envs/seq2science/bin/seq2science", line 213, in _run
    raise ValueError("Please specify at least 12 cores.")
ValueError: Please specify at least 12 cores.
```

**To Reproduce**

```
seq2science run atac_seq --cores 4
```
**Expected behavior**

Run with less cores, but slower.

Maarten-vd-Sande:
This is an "issue" with how snakemake is incapable of scaling pipes to lower threads (alignment 10 threads + 2 threads sorting). I can see how this is unexpected behaviour, and the solution should be easy.

PR #402, opened 2020-06-12
Flexible cores, issue #391

Maarten-vd-Sande:
We do the flexible core parsing in pipes for snakemake, since snakemake does not support this.

### DIFF
--- a/bin/seq2science
+++ b/bin/seq2science
@@ -312,6 +312,7 @@ def core_parser(cores):
         try:
             raise argparse.ArgumentError(core_arg, "specify at least two cores.")
         except argparse.ArgumentError as e:
+            print()  # empty line
             print(e)
             sys.exit(1)
 
--- a/bin/seq2science
+++ b/bin/seq2science
@@ -112,7 +112,9 @@ def seq2science_parser(workflows_dir="./seq2science/workflows/"):
     run.add_argument(
         "-j", "--cores",
         metavar="N",
-        help="Use at most N cores in parallel. Requires at least 12 cores.",
+        type=int,
+        required=True,
+        help="Use at most N cores in parallel.",
     )
     run.add_argument(
         "-n", "--dryrun",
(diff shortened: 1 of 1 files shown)


Maarten-vd-Sande:
Yeah I guess you are right, we should inform the user. I do not like a default, especially not a low default of 2. I am afraid lots of people will never change their default and are unnecessarily waiting for a long time since cores has been set too low.

Maarten-vd-Sande:
Okay I set it so that you need to specify 2 cores or more, and otherwise it will print a nice error message telling you to use 2 or more. So no unexpected behaviours :)

```
$ seq2science run alignment --cores 1 -c config.yaml
usage: seq2science run [-h] [-c FILE] -j N [-n] [--unlock]
                       [--snakemakeOptions KEY=VAL [KEY=VAL ...]]
                       [--profile PROFILE NAME] [-k] [--rerun-incomplete]
                       {chip_seq,scatac_seq,alignment,download_fastq,rna_seq,atac_seq}

Run a complete workflow. This requires that a config and samples file are
either present in the current directory, or passed as an argument.

positional arguments:
  {chip_seq,scatac_seq,alignment,download_fastq,rna_seq,atac_seq}

optional arguments:
  -h, --help            show this help message and exit
  -c FILE, --configfile FILE
                        The path to the config file.
  -j N, --cores N       Use at most N cores in parallel. Must be at least 2.
  -n, --dryrun          Do not execute anything, and display what would be
                        done.
  --unlock              Remove a lock on the working directory.
  --snakemakeOptions KEY=VAL [KEY=VAL ...]
                        Extra arguments to pass along to snakemake. An example
                        could be seq2science run alignment --cores 12
                        --snakemakeOptions resources={mem_gb:100}
                        local_cores=3. Here we pass local_cores as KEY=VALUE
                        and additional resources can even be passed along in a
                        dictionary. Take a look at the snakemake API for a
                        complete list of all possible options: https://snakema
                        ke.readthedocs.io/en/latest/api_reference/snakemake.ht
                        ml
  --profile PROFILE NAME
                        Use a snakemake/seq2science profile. Profiles can be
                        taken from: https://github.com/snakemake-profiles
  -k, --keep-going      Go on with independent jobs if a job fails.
  --rerun-incomplete    Re-run all jobs the output of which is recognized as
                        incomplete.

argument -j/--cores: specify at least two cores.
```

PR #403, opened 2020-06-12
ucsc trackhubs are now pickled, to prevent repeated reruns.

### LINKED ISSUE #398: Q: track hubs and UCSC genomes
If I'm correct, `seq2science` generates files for a track hub, even if the genome comes from UCSC. However, if a genome is on UCSC, all the files will already be there. In that case you can just use the UCSC browser right? This would even be preferred as UCSC will contain loads of additional tracks and information.

siebrenf:
fixes #398 by pickling the assemblies

this way we can drop the dryrun variable, which was only used here anyway, and use a more robust approach to trackhub generation.

It should be parallelizable from the box.


Maarten-vd-Sande:
(one advantage of FileLock is that it **should** remove the lock when the program is interrupted or stopped)

PR #409, opened 2020-06-16
use mamba in snakemake too!

siebrenf:
the tests werent actually using mamba. sloooow...


Maarten-vd-Sande:
Just FYI the logo shift is/was only in the tests

Maarten-vd-Sande:
It still happens in the tests btw

siebrenf:
😲

siebrenf:
at least the mamba fix works :)

PR #411, opened 2020-06-22
Peak counts

Maarten-vd-Sande:
Update the way the count table for ATAC-seq & ChIP-seq is made. These count tables can be directly loaded into gimme maelstrom for instance, and thus opening the way for us to start supporting gimme :tada:.
Supports: 
* quantile norm
* TMM
* RLE
* upperquartile

**What did change**
Peaks are merged via combine_peaks and all made the same width. Then fed through the different normalizers.


Maarten-vd-Sande:
I am running it on a big dataset now to see if everything looks okay, so no merging yet :)

seq2science/rules/peak_counts.smk
```diff
@@ -1,48 +1,288 @@
...
+
+def count_table_output():
+    """
+    get all the count table outputs.
+    """
+    ftype = get_ftype(list(config['peak_caller'].keys())[0])
+    if ftype != "narrowPeak":
+        return []
+
+    print(expand(["{result_dir}/count_table/{peak_caller}/{assemblies}_{normalization}.tsv"],
```
siebrenf, line 12: smart testing method. but can be removed ofcourse.

additionally, you are `expand`ing with the config, so could you not replace all those f"...{config['logbase]" with "...{logbase}"?

Maarten-vd-Sande: Ouch, I missed that print! It's huge, difficult to miss :blush: 

No this can not be replaced, unfortunately. There are two string "interpolations" happening at the same time. Once in the normalization kwargs, and once in the path string. :(

siebrenf: ah OK, and an expand(expand()) isn't possible I guess?

Maarten-vd-Sande: Probably possible, but then you would have to do some curly braces magic which I think is a worse solution for the next person touching that... :thinking:

Maarten-vd-Sande:
Just a little confirmation for myself:

**raw** counts:
image

**quantile normalized** counts:
image

**TMM** counts:
image

**RLE** counts:
image

**upper quartile** counts:
image

Maarten-vd-Sande:
random peak:
```
        samples... ->
raw   84     68     102    94
qno   5.23   3.95   5.71   5.13
TMM   5.19   3.84   5.52   4.94  
RLE   5.27   3.91   5.55   5.01  
uqn   5.20   3.96   5.76   5.02  
```
good enough for me!

PR #412, opened 2020-06-25
Automatic detection of annotation without checkpoints

Maarten-vd-Sande:
Now we should be able to automatically detect whether or not an assembly has an annotation :tada:, no stupid if statement required + other pipelines' trackhub now can also get annotations :).

**What did change**
Function that queries genomepy for an annotation of an assembly. Should the function crash or return false when no genome has been found?


seq2science/rules/get_genome.smk
```diff
@@ -118,3 +127,31 @@ rule decoy_transcripts:
...
+def has_annotation(assembly):
+    """
+    returns True/False on whether or not the assembly has an annotation.
+    """
+    # check if genome providd by user or already downloaded, if so check if the annotation came along
+    if all(os.path.exists(f"{config['genome_dir']}/{assembly}.{extension}") for extension in config["genome_types"]):
+        return os.path.exists(f"{config['genome_dir']}/{assembly}.annotation.gtf")
+
+    if "provider" in config:
+        providers = config["provider"]
```
siebrenf, line 142: should this not be a list?

`providers = [config["provider"]]`

Maarten-vd-Sande: correct

seq2science/rules/get_genome.smk
```diff
@@ -118,3 +127,31 @@ rule decoy_transcripts:
...
+        providers = ["Ensembl", "UCSC", "NCBI"]
+
+    # check if we expect an annotation
+    # we do not want genomepy outputting to us that it is downloading stuff
+    with open(os.devnull, 'w') as null:
+        with contextlib.redirect_stdout(null), contextlib.redirect_stderr(null):
+            for provider in providers:
+                p = genomepy.ProviderBase.create(provider)
+                if assembly in p.genomes and \
+                    p.get_genome_download_link(assembly) is not None:
```
siebrenf, line 153: isn't this double? If the assembly is in p.genomes, there should always be a genome_download_link.

Maarten-vd-Sande: I guess? I don't know how genomepy works internally :). I had some issues with just using get_genome_download_link, and added the check if its in genomes. I didn't make the effort to check if it's superfluous. :angel:

Maarten-vd-Sande:
Tnx for thorough feedback :)

Maarten-vd-Sande:
So you mean to log a warning if no annotation is found? That makes sense, I will add it.

PR #415, opened 2020-06-29
Black

siebrenf:
black formatted:
- alignment.smk
- peak_calling.smk
- quantifiaction.smk

still not formatted:
- config_generic.smk
- qc.smk


siebrenf:
`snakefmt` seems to have some issues with indenting. In some cases (that I cannot clearly define), it removes too much indenting, breaking the code...

The last 2 files throw errors that I cannot interpret :(

PR #423, opened 2020-07-03
cleanup and qnorm

Maarten-vd-Sande:
The original idea was that this PR would contain differential peak calling for atac/chip seek. But unfortunately, I do not have the time to implement this at the moment. Changes:

- cleanup
- quantile normalizaiton from package
- give onehot encodings of which peaks can be found in which samples

Somewhere in the future differential peaks might still be implemented. Would be nice, especially feeding its output to gimme.
https://support.bioconductor.org/p/132840/


Maarten-vd-Sande:
https://support.bioconductor.org/p/132840/

seq2science/scripts/qnorm.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+df = pd.read_csv(snakemake.input[0], comment="#", index_col=0, sep="\t")
+
+# cpm normalization
+df = df * 1_000_000 / df.sum(axis=0)
+
+# quantile normalize
+df_qn = qnorm.quantile_normalize(df)
+open(str(snakemake.output[0]), "w").write(
```
siebrenf, line 12: nitpick: Since you named the output arguments, I would use those names here for clarity.
IIRC something like `snakemake.output$real`

seq2science/rules/peak_counts.smk
```diff
@@ -301,3 +268,43 @@ rule mean_center:
...
+        """
+        awk '{{print $1":"$2"-"$3}}' {input.combinedpeaks} > {output.real} 2> {log}
+
+        for brep in {input.narrowpeaks}
+        do
+            bedtools sort -i $brep 2> {log} |
+            bedtools intersect -a {input.combinedpeaks} -b stdin -c 2> {log} |
+            awk '{{print $4}}' 2> {log} |
+            paste {output.real} - > {output.tmp}
+            mv {output.tmp} {output.real}
```
siebrenf, line 306: If I understand correctly, you are trying to append each `brep` to the output file. I think you could do that more simply with `awk '{{print $4}}' >> {output.real} 2> {log}` (`>>` creates a file if missing, and appends if existing)

Furthermore, you could start with the echo below. That way you can omit the use of `output.tmp`. If an error occurs, snakemake will delete the file anyway, so no need to add the comment only after the loop.

Maarten-vd-Sande: It adds columns, not rows, you end up with a file like this:

```
    sample1    sample2    sample3
peak1    0    1    0
peak2    1    1    1
peak3    0    0    1

```

seq2science/scripts/qnorm.py
Maarten-vd-Sande: It's not a named output I believe, otherwise I would very much agree

seq2science/rules/peak_counts.smk
siebrenf: aah. Thanks!

PR #424, opened 2020-07-07
Cite func

Maarten-vd-Sande:
Added functionality:

```
maarten@lap-maarten:~/Desktop/seq2science$ seq2science explain atac_seq -c config.yaml

Preprocessing of the reads was done automatically with workflow tool seq2science v0.0.3 (https://doi.org/10.5281
/zenodo.3921913). Public samples were downloaded from the Sequence Read Archive (https://doi.org/10.1093/nar/gkq1019) 
with help of the ncbi e-utilities. Genomic assembly Bl71nemr was downloaded with GenomePy v0.8.3 (https://doi.org/10.21105
/joss.00320). We trimmed paired-end reads with trim galore! v0.6.4 (http://www.bioinformatics.babraham.ac.uk/projects
/trim_galore/) with options '--quality 10 --length 20 --consider_already_trimmed 10' and cutadapt (https://doi.org/10.14806
/ej.17.1.200). Fastq quality was measured by FastQC v0.11.8. Reads were aligned with bwa-mem v0.7.17 (http://arxiv.org
/abs/1303.3997) with options '-M'. Mapped reads were removed if they did not have a minimum mapping quality of 30, were a 
(secondary) multimapper, aligned inside the ENCODE blacklist (https://doi.org/10.1038/s41598-019-45839-z), and finally were 
tn5 bias shifted by seq2science. General alignment statistics were collected by samtools stats v1.10. Afterwards duplicate reads 
were marked with picard MarkDuplicates v2.21.2. Peaks were called with macs2 v2.2.7 with options '--shift -100 --extsize 200 
--nomodel --keep-dup 1 --buffer-size 10000' in BAM mode. Narrowpeak files of replicates belonging to the same condition were 
merged with fisher's method in macs2 v2.2.7. The fraction reads in peak score (frips) was calculated by featurecounts v1.6.4 
(https://doi.org/10.1093/bioinformatics/btt656). We used the UCSC genome browser (http://www.genome.org/cgi/doi/10.1101
/gr.229102) to visualize and inspect alignment. Quality control metrics were aggregated by MultiQC v1.9.
```

And moved the cli to cli.py

@siebrenf can you do this for RNA-seq?

**What did change**
I added a function called explain_rule. You wrap the explanation in message with this function and seq2science handles the rest for you.

Example:

```
rule call_peak_genrich:
    """
    Call peaks with genrich based on the pileup.
    """
    input:
        log=expand("{result_dir}/genrich/{{fname}}.log", **config),
    output:
        narrowpeak=expand("{result_dir}/genrich/{{fname}}_peaks.narrowPeak", **config),
    log:
        expand("{log_dir}/call_peak_genrich/{{fname}}_peak.log", **config),
    benchmark:
        expand("{benchmark_dir}/call_peak_genrich/{{fname}}.benchmark.txt", **config)[0]
    message: explain_rule("""
        Peaks were called with genrich v@genrich[genrich] 
        (https://github.com/jsh58/Genrich) with options '{config[peak_caller][genrich]}'.
        """)
    conda:
        "../envs/genrich.yaml"
    params:
        config["peak_caller"].get("genrich", ""),
    shell:
        """
        Genrich -P -f {input.log} -o {output.narrowpeak} {params} -v > {log} 2>&1
        """
```
The @ indicates which file, and the [] which package inside that file.


seq2science/rules/configuration_generic.smk
```diff
@@ -42,6 +42,9 @@ docs: https://vanheeringen-lab.github.io/seq2science
 """
 )
 
+include: f"{config['rule_dir']}/explain.smk"
```
siebrenf, line 45: nice way to minimize repeating code.

we could also add other universal rules at the bottom of this file to clean up the snakefiles. Personally, I would only add the first of these two however.
```
include: f"{config['rule_dir']}/configuration_workflows.smk"
include: f"{config['rule_dir']}/get_fastq.smk"
```

seq2science/rules/alignment.smk
```diff
@@ -48,8 +48,9 @@ if config["aligner"] == "bowtie2":
             pipe(expand("{result_dir}/{aligner}/{{assembly}}-{{sample}}.samtools-coordinate.pipe", **config)[0]),
         log:
             expand("{log_dir}/bowtie2_align/{{assembly}}-{{sample}}.log", **config),
-        group:
-            "alignment"
```
siebrenf, line 52: what's with `group: alignment` being removed?

Maarten-vd-Sande: Originally this was necessary for rules to be executed at the same time. But since we use pipes it is not necessary, and seems to do nothing anymore. So I decided to just remove it and replace with message

seq2science/rules/configuration_generic.smk
Maarten-vd-Sande: This does not seem to work because the order in which these files are read is important :(

PR #426, opened 2020-07-10
set workflow name from _ to -

qxuchn:
Change the **workflow** names, and let the name more accepable.

**What did change**
rna_seq,download_fastq,alignment,scatac_seq,atac_seq,chip_seq
to
rna-seq,download-fastq,alignment,scatac-seq,atac-seq,chip-seq

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,7 +2,10 @@
 A description of the problem that is being solved, or which new feature is implemented. Please link to other relevant issues/pr if relevant
 
 **What did change**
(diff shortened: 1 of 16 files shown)


Maarten-vd-Sande:
Let me know if you need help finishing the checklist :)

@siebrenf I don't think there are any real obligations against this right? Seems better to me

Maarten-vd-Sande:
@qxuchn once you update the docs and changelog we can merge this :smile:

Issue #333, opened 2020-05-23
FR: support bwa mem2


Maarten-vd-Sande:
Quite self-explanatory. Should be much faster -> much cooler.

Maarten-vd-Sande:
Still a WIP according to their github page, _not recommended for production uses at the moment_. Will implement it once "done". Should be twice as fast.
https://github.com/bwa-mem2/bwa-mem2

Maarten-vd-Sande:
> Indexing the reference sequence (Requires 28N GB memory where N is the size of the reference sequence).

That's a yikes

PR #428, opened 2020-07-11
Bwamem2

Maarten-vd-Sande:
Also #333. I think we are the first workflow to support bwa-mem2!

**What did change**
Simply added another aligner in alignment.smk

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
No problems when doing a couple samples on cn45 :tada:

PR #431, opened 2020-07-15
counts from bams

siebrenf:
STAR was the only aligner accepted by the RNA-seq workflow. STAR produced counts directly. Further filtering applied to the bam files downstream did therefore not apply to the counts.

**What did change**
Implemented `HTSeq-count` (default) and `featureCounts`. Both summarize the gene counts from the filtered and marked bam files.
STAR and HISAT2 are now suitable for use in RNA-seq. An error will warn the user when attempting to run with another aligner.

**Checklist**
- [x] I updated the CHANGELOG


siebrenf:
It actually was! A nice suprise :)

Just a few more tweaks then the workflow is up to date!

Maarten-vd-Sande:
just a little fyi if you are bored, you could add the images to the multiqc report if you want
https://multiqc.info/docs/#images

siebrenf:
> Note that this means that it's very possible to make the HTML file very very large if abused!

5 images are generated per design with a batch effect. I think I'll spend my time on the `explain` feature ;)

PR #433, opened 2020-07-16
explain now stores data in a dict in explain.smk

siebrenf:
rules are bulky. explain is nice, but doesnt help readability

**What did change**
- moved as much of the explain feature as possible to `explain.smk` to keep the rules clean
- add RNA-seq workflow explain features

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/cli.py
+++ b/seq2science/cli.py
@@ -11,6 +11,7 @@ import re
 import inspect
 import yaml
 import contextlib
+from collections import OrderedDict
 
 
 # we need to be able to get the parser from the file without a valid seq2science installation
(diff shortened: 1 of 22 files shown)


seq2science/cli.py
```diff
@@ -277,6 +278,8 @@ def _explain(args, base_dir, workflows_dir, config_path):
 
     rules_used = {"start": f"\nPreprocessing of reads was done automatically with workflow tool "
                            f"seq2science v{seq2science.__version__} (https://doi.org/10.5281/zenodo.3921913)."}
+    rules_used = OrderedDict(rules_used)
+
```
Maarten-vd-Sande, line 282: Just fyi. Python 3.6+ dictionaries are always insertiond ordered. So this actually does not change anything

siebrenf: right, I see the variability of the output is in the DAG. It changes with and without the orderedDict...

PR #434, opened 2020-07-17
zenodo JSON

siebrenf:
The zenodo page of s2s looks like a mess. According to [this page](https://zenodo.org/account/settings/github/repository/vanheeringen-lab/seq2science#), we can add this metadata file to specify everything.

However! Documentation on the metadata is sketchy. Best I found so far is [this article](https://zenodo.org/record/3531504). No idea (yet) if we can leave fields empty, to be filled automatically (like the version number).

### DIFF
--- a/.github/workflows/continuousdeployment.yml
+++ b/.github/workflows/continuousdeployment.yml
@@ -17,6 +17,13 @@ jobs:
       with:
         python-version: 3.6
 
+    - name: Get Changelog Entry
+      id: changelog_reader
+      uses: mindsers/changelog-reader-action@latest
+      with:
+        version: ${{ github.ref }}
+        path: ./CHANGELOG.md
+
     - name: Create Release
       uses: actions/create-release@latest
       env:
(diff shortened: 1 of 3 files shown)


Maarten-vd-Sande:
The tests are getting a bit unstable :S

siebrenf:
Jenkins has been getting stuck (not progressing, but not crashing either) at several places. Local tests work fine though...

But aren't you supposed to be on holiday now? :D

Maarten-vd-Sande:
I'm leaving tomorrow :palm_tree: 

My first guess is that it has something to do with the annotation function:
https://github.com/vanheeringen-lab/seq2science/blob/5c2f117f600ae6e335b36dd3523822fcd46423ac/seq2science/rules/get_genome.smk#L138-L170

Since it is hanging in the making of the DAG, and this happens during the making of the dag and is a relatively new addition. You should be able to login to Jenkins and cancel/restart those tests right?

siebrenf:
I already did a few times yesterday. It hangs in several places, but ill
look into the annotation function. Good tip!

siebrenf:
I think it cannot operate in parallel, and freezes when two tests are run simultaneously

Maarten-vd-Sande:
Ugh... I can not really think of anything else that causes this 🤔. Since it happens after the print "generating DAG". 

One wat to pinpoint if it is this function or not is to add a print at the start and the end of this function, to see if it hangs in between those two prints. Also a good idea to turn off the contextmanager, to see if anything useful from genomepy gets printed.

Don't feel pressured into finding this out, half of the reason I am writing this is that it's a good reminder for me ☺️

siebrenf:
[json explanations](https://developers.zenodo.org/#representation)
[zenodo's json schema](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/deposit/jsonschemas/deposits/records/record-v1.0.0.json)

siebrenf:
I updated the deployment action to automatically read the changelog, then updated the changelog to follow the correct format.
The `.zenodo.json` adds the major contributors to the "creators" list, and the other contributors to the "contributors" list. Since the latter requires the field "type", I set this to "Project Member" (best alternative would be "Other", but that looks weird IMO).

**What this will do**
- Each github release will have the 1-line abstract of our workflow + the latest changelog entry
- Zenodo with reflect that, plus all authors and contributors. [Test page (lacking the changelog entry).](https://sandbox.zenodo.org/record/652097)

PR #435, opened 2020-07-21
minor fixes

siebrenf:
small stuff


siebrenf:
manually cancelling and rerunning the test twice to check if a mutex lock on the genomepy cache does the trick

Maarten-vd-Sande:
Nice 😄

PR #438, opened 2020-07-23
skip missing contigs in trackhub_index

### LINKED ISSUE #436: BUG: Indexing of Spur_5.0 assembly fails
**Describe the bug**
When making an index for assembly Spur_5.0 rule [trackhub_index](https://github.com/vanheeringen-lab/seq2science/blob/5c2f117f600ae6e335b36dd3523822fcd46423ac/seq2science/rules/trackhub.smk#L163) fails:
`NC_001453.1 is not found in chromosome sizes file`

**To Reproduce**
I ran the atac-seq pipeline with many samples, but I guess this should be enough to reproduce:

```
sample              assembly
GSM2546188   Spur_5.0
``` 

**Expected behavior**
No failure! Is this a genomepy+annotation thingy? Or is it on our side?

siebrenf:
fix #436 (probably)


Maarten-vd-Sande:
I'll check this one once I'm back :)

Maarten-vd-Sande:
Unfortunately this does not solve the problem :sob:

Maarten-vd-Sande:
@siebrenf I can make a tmp file that only contains the gtf stuff that is in the .fa.sizes file. Do you expect any problems with that?

siebrenf:
OK, as we saw, this is an issue with the current Spur_5.0 assembly and annotation.
IMO the PR can still be merged for situations where there's one a few contigs missing.

PR #440, opened 2020-07-23
RNA-seq: strandedness

siebrenf:
Most RNA-seq protocols nowadays are strand-specific, but finding out which protocol was used, and what this means for gene quantification is a hassle.

**What did change**
Using RSeQC, all RNA-seq samples for which the strandedness is unknown will be tested. 

These tests can be omitted by either filling in the `strandedness` column in the `samples.tsv` (test will only be done for missing fields), or by setting `ignore_strandedness` in the `config.yaml`.

This knowledge is used by s2s to fine-tune gene quantification (HTSeq and featureCounts), and to split bigwigs by strand.
RSeQC results are incorporated in the MultiQC.

**Checklist**
- [x] I updated the CHANGELOG


seq2science/rules/configuration_workflows.smk
```diff
@@ -67,6 +67,18 @@ for conf_dict in ["aligner", "quantifier", "diffexp"]:
...
+# ...for rna-seq
+if get_workflow() == "rna_seq":
+    # delete the old strandedness report if samples.tsv was updated
+    strandedness_report = f"{config['qc_dir']}/strandedness/inferred_strandedness.tsv"
+    if os.path.exists(strandedness_report) and not config['ignore_strandedness']:
+        strandedness = pd.read_csv(strandedness_report, sep='\t', dtype='str', index_col=0)
+        col = samples.replicate if "replicate" in samples else samples.index
+        if len(strandedness.index) != len(set(col)) or not all(s in set(col) for s in strandedness.index):
+            os.unlink(strandedness_report)
+
```
Maarten-vd-Sande, line 80: This is a bit iffy.. It breaks the flow of the rest of the snakemake code, by doing this with python. Currently it just sucks a bit the way snakemake resolves what needs to be run, so I am fine with it. 

Something for a blue monday in the far future: we could consider doing this (and other file dependencies) more properly. Not entirely sure what would be a good way to do it, but for instance checksumming the config and samples file, and using those checksums as inputs for rules or even some config options. However that is going to make the rules even more bulky, will introduce weird bugs, and is a lot of work.

docs/content/workflows/atac_seq.md
```diff
@@ -7,7 +7,7 @@ Running an ATAC-seq analysis has never been easier!
 </p>
 
 #### Downloading of sample(s)
-Depending on whether the samples you start seq2science with is your own data, public data, or a mix, the pipeline might start with downloading samples. Take a look at the [downloading_fastq](download_fastq.md) workflow for extensive documentation about downloading of public samples. 
+Depending on whether the samples you start seq2science with is your own data, public data, or a mix, the pipeline might start with downloading samples. Take a look at the [downloading_fastq](https://vanheeringen-lab.github.io/seq2science/content/workflows/download_fastq.html) workflow for extensive documentation about downloading of public samples.
```
Maarten-vd-Sande, line 10: thanks for fixing, I thought this way of linking worked!

PR #443, opened 2020-07-30
DEXSeq

siebrenf:
You can now get the DEXSeq counts matrix from the RNA-seq workflow by setting `dexseq: True` in the config.yaml.

**Feature creep**
OK, so I did a few more things that should have been their own PR :( 
The significant changes are all in `cli.py`
- default cores set to 2. Why? Because i dont like specifying cores on a dryrun.
- added the `-r/--reason` flag to the s2s CLI, mirroring snakemake. Why? I use it a lot, and am lazy :)
- improved CLI error feedback for dictionaries (such as `config={a:b}`)

**Checklist**
- [x] I made a PR to ~~develop~~ a branch (not master)
- [x] I updated the CHANGELOG


CHANGELOG.md
```diff
@@ -13,20 +13,24 @@ All changed fall under either one of these types: `added`, `changed`, `deprecate
...
-- rules and scriptnames in RNA-seq. ex: `txi.R` is now `quant_to_counts.R` to better reflect its function
+- rules and script names in RNA-seq. ex: `txi.R` is now `quant_to_counts.R` to better reflect its function
 - `quant_to_counts.R` now converts salmon transcript abundances to gene counts identically to DESeq2
 - STAR no longer outputs counts, and is no longer found under `quantifiers`
 - gene counts are generated from (filtered) bams when using either STAR or HISAT2 as aligner and HTSeq or featureCounts are quantifier
 - batch corrected gene counts are generated if a DESeq2 design contrast inclused a batch
 - batch corrected TPM are generated if a DESeq2 design contrast inclused a batch, and quantification was performed using Salmon
   - for us in ANANSE, for instance
 - `seq2science explain` now retrieves messages from `explain.smk`.
+- seq2science CLI now has default cores set to 2 (-j/--cores flag)
```
Maarten-vd-Sande, line 33: I think this is a bad idea. I want people to consider the number of cores they want to use, and when setting it to 2, I am afraid lots of people will be waiting unnecessarily long. Why don't we set it to 2 (or any number) for a dry run, but for the real run people still have to set it?

Snakemake also switched that you HAVE to supply the nr of cores when running, so I guess they also had the same problem.

seq2science/rules/gene_counts.smk
```diff
@@ -108,3 +108,47 @@ else:
 
                 counts.index.name = "gene"
                 counts.to_csv(output[0], sep="\t")
+
+
+if config.get("dexseq"):
```
Maarten-vd-Sande, line 113: ah didnt know this works :), looks cleaner than the get.("key", False) thing

seq2science/cli.py
```diff
@@ -237,15 +243,33 @@ def _run(args, base_dir, workflows_dir, config_path):
...
+            
+    # scale threads if cores is too low
+    overwrite_threads = core_parser(parsed_args)
+    if overwrite_threads:
+        if "overwrite_threads" not in parsed_args:
+            parsed_args["overwrite_threads"] = overwrite_threads
+        else:
+            for k, v in overwrite_threads.items():
+                if k not in parsed_args["overwrite_threads"]:
+                    parsed_args["overwrite_threads"][k] = v
```
Maarten-vd-Sande, line 272: Just an idea, maybe cleaner if we move this into the core_parser function?

PR #459, opened 2020-08-07
Ena download

Maarten-vd-Sande:
Directly downloading from ENA has the advantage that we can skip the dump step! Does NOT support aspera atm though..

**What did change**

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
Sjonge jonge, again scRNA came along?!

seq2science/rules/configuration_generic.smk
```diff
@@ -358,6 +365,67 @@ for sample in samples.index:
                          f"access is currently not supported. We advise you to download the sample "
                          f"manually, and continue the pipeline from there on.")
 
+trace_tp.close()
+eutils_tp.close()
+
+def url_is_alive(url):
```
siebrenf, line 371: hope you don't mind some unasked advise:

I've made a similar function in genomepy, and noticed certain websites get hung up with too many requests. This function seemed to be fairly robust for me:
```
def check_url(url, time_out=3, max_tries=1, _try=1):
    """Check if URL works. Returns bool"""
    try:
        ret = urllib.request.urlopen(url, timeout=time_out)
        # check return code for http(s) urls
        if url.startswith("ftp") or ret.getcode() == 200:
            return True

    except (urllib.request.URLError, socket.timeout):
        # Some providers get swamped with requests, so we allow retries
        if _try <= max_tries:
            time.sleep(1)
            return check_url(url, time_out=time_out, max_tries=max_tries, _try=_try + 1)
        return False
```

Maarten-vd-Sande: Yes, I was planning to implement something like this.

Why did you make this recursive and not just sequential?

siebrenf: so you can set the maximum number of tries as a variable. The code stays clean even if you want to give a URL 10 tries.

Maarten-vd-Sande: Wouldn't this work?
```
for try in range(max_tries):
    try:
        ret = urllib.request.urlopen(url, timeout=time_out)
        # check return code for http(s) urls
        if url.startswith("ftp") or ret.getcode() == 200:
            return True
    except (urllib.request.URLError, socket.timeout):
        continue
    return False
```

Maarten-vd-Sande: Not saying either is better, but I usually struggle with implementing recursive stuff :P

siebrenf: yes it would. this just popped in my head faster

don't forget to add a pause in the except though!

Maarten-vd-Sande:
Now testing on a large dataset and alignment:

```
Job counts:
        count   jobs
        1       align_all
        8       bwa_index
        199     bwa_mem
        8       complement_blacklist
        104     ena2fastq_PE
        2       ena2fastq_SE
        8       get_genome
        93      id2sra
        199     mark_duplicates
        199     samtools_index
        199     samtools_presort
        8       setup_blacklist
        199     sieve_bam
        90      sra2fastq_PE
        3       sra2fastq_SE
        194     trim_galore_PE
        5       trim_galore_SE
        1519
```

Maarten-vd-Sande:
Everything works just fine for "my" samples.

Will test if it resolves some issues with difficult samples of Rebecca

Maarten-vd-Sande:
I downloaded the 300 samples of Rebecca (500GB) in 5 hours with aspera, so an average of 25 mb/s. Not bad :)

seq2science/rules/configuration_generic.smk
```diff
@@ -300,7 +311,7 @@ with FileLock(layout_cachefile_lock):
         layout_cache = {}
 
 
-    trace_tp = ThreadPool(20)
+    trace_tp = ThreadPool(40)
```
siebrenf, line 314: should this be `config["threads"]`?

Maarten-vd-Sande: Well Python's threads do not really work well in the sense that they won't be executed in "parallel". 40 threads together will never use more than 1 cpu thread. However when threads are waiting for stuff (e.g. waiting for websites to reply), non-waiting threads can continue with stuff. So the threads here are in a sense not the same as the threads of Snakemake. 

40 might be a bit of overkill though, 20 should be fine I think, I'll revert that back.

PR #462, opened 2020-08-07
fixed profiles

### LINKED ISSUE #460: BUG: profile parsed incorrectly
**Describe the bug**
Now it parsed wrongly:
`config                 : ['email=M.vanderSande@science.ru.nl', 'ascp_path=$HOME/.aspera/connect/bin/ascp', 'ascp_key=$HOME/.aspera/connect/etc/asperaweb_id_dsa.openssh']`

siebrenf:
Fix #460 

**What did change**
the profiles we use have dictionaries that are phrased as lists:
```
config:
- threads=12
- genome_dir=$HOME/.local/genomes
```

~~This PR adds a parser that can detect this silly notation of ours, and converts it without overwriting CLI input. Its only while writing this that I realize I should instead have changed how we write profiles (using actual dict syntax). Especially since s2s arguments now must be aimed at the API, rather than the CLI...~~

This notation is incorrect. Ill change the cookiecutter. The cookiecutter will also use s2s compatible (API) names.

This PR adds a parser so the profile config no longer overwrites CLI input

**Checklist**
- [x] I updated the CHANGELOG


seq2science/cli.py
```diff
@@ -201,6 +203,38 @@ def _init(args, workflows_dir, config_path):
...
+    """read profile and add new arguments to parsed args"""
+    profile = yaml.safe_load(open(profile_file).read())
+
+    # parse yaml dicts
+    for k, v in profile.items():
+        if isinstance(v, list) and "=" in v[0]:
+            profile[k] = {}
+            for e in v:
+                k2, v2 = e.split("=")
+                profile[k][k2] = v2
```
siebrenf, line 225: This section converts a YAML list with `=` in the elements to a dict. Without this section, the code works as intended with proper profile YAMLs.

siebrenf:
Other than our YAML format, this code actually does something that needed to be done: it parses a profile YAML without overwriting the CLI arguments.

So @Maarten-vd-Sande, do we want to support the snakemake-profiles as they are (and keep the above mentioned section), or should they be tweaked to create actual YAML dicts?

siebrenf:
I think we should launch a s2s-profiles, as the profile contents must be aimed at the API now (e.g. `keep_going` must become `keepgoing`). There we can fix the YAML notation too

Maarten-vd-Sande:
I think I am not following. 

I think we should solve it the same way as snakemake does:
https://github.com/snakemake/snakemake/blob/master/snakemake/__init__.py#L850

I think I originally tried doing something like this but it got lost in the transition.

siebrenf:
This PR is now functional with proper profile YAMLs, such as this test config in `~/.config/snakemake/test/config.yaml`:

```
cores: 48
config:
  ascp_path: $HOME/.aspera/connect/bin/ascp
  ascp_key: $HOME/.aspera/connect/etc/asperaweb_id_dsa.openssh
keepgoing: True
```

previous snakemake-profile YAML are **not compatible**:
```
cores: 48
config:
  - ascp_path=$HOME/.aspera/connect/bin/ascp
  - ascp_key=$HOME/.aspera/connect/etc/asperaweb_id_dsa.openssh
keep_going: True
```

As discussed I'll change the cookiecutter next :)

PR #465, opened 2020-08-10
Fasterq dump

Maarten-vd-Sande:
Switched to fasterq-dump. Should be faster according to random benchmark. Makes sense because parallel-fastq-dump is just fastq-dump in parallel vs an implemenatataion that is made for parallel.

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
parallel-fastq dump makes use of $TMPDIR, but I believe that fasterq-dump makes use of some weird config file of them. TODO: check if this works!

Maarten-vd-Sande:
Not sure if it's faster, but at least we switched to the newer fastq-dump / actively maintained fastq-dump.

seq2science/rules/get_fastq.smk
```diff
@@ -176,17 +183,24 @@ rule sra2fastq_PE:
...
-        parallel-fastq-dump -s {input}/* -O $tmpdir {config[split]} \
-        --threads {threads} --gzip >> {log} 2>&1
+        # acquire the lock
+        (
+            flock --timeout 30 200 || exit 1
+            sleep 2
+        ) 200>{layout_cachefile_lock}
+
+        # dump
+        fasterq-dump -s {input}/* -O $tmpdir --threads {threads} --split-3 >> {log} 2>&1
```
siebrenf, line 193: out and tmp directories should be on the same disk, something like this?

```
# setup tmp dirs
out={config[sra_dir]}/out/{wildcards.sample}
mkdir -p $out; trap "rm -rf $out" EXIT
tmp={config[sra_dir]}/tmp/{wildcards.sample}
mkdir -p $tmp; trap "rm -rf $tmp" EXIT

...

fasterq-dump -s {input}/* -O $out -t $tmp --threads {threads} --split-3 >> {log} 2>&1

...

for f in $(ls -1q $out | grep -oP "^[^_]+" | uniq); do
```

siebrenf, line 193: nitpick: --split-3 seems the default for `fasterq-dump`

seq2science/schemas/config/download.schema.yaml
```diff
@@ -28,13 +28,3 @@ properties:
...
-
-  # fastq-dump parameters
-  split:
-    description: fastq-dump parameters when splitting reads (paired-end)
-    default: --split-files --skip-technical --dumpbase --readids --clip --read-filter pass --defline-seq '@$ac.$si.$sg/$ri' --defline-qual '+'
-    type: string
-  splot:
-    description: fastq-dump parameters when splotting reads (single-end)
-    default: --split-spot  --skip-technical --dumpbase --readids --clip --read-filter pass --defline-seq '@$ac.$si.$sg/$ri' --defline-qual '+'
-    type: string
```
siebrenf, line 40: question: so these are all done by default now? Their wiki only mentioned split-spot and skip-technical.

siebrenf (changes requested):
For fasterq-dump, the tmp dir is the current directory by default. I think we want this to be on the same location as the out dir.

seq2science/schemas/config/download.schema.yaml
Maarten-vd-Sande: Yes, fasterq-dump has "reasonable" defaults apparantly

seq2science/rules/get_fastq.smk
Maarten-vd-Sande: I know, I just thought this is more clear in what the difference is between those two

siebrenf:
While browsing their wiki I also found [this ascp warning](https://github.com/ncbi/sra-tools/wiki/Avoid-using-ascp-directly-for-downloads). Should we still support ASCP?

seq2science/rules/get_fastq.smk
Maarten-vd-Sande: Yeah that would make sense

siebrenf: 👍

seq2science/schemas/config/download.schema.yaml
siebrenf: nice! just wanted to check :)

Maarten-vd-Sande:
Oh that is interesting! I didnt know about the ascp thing!! I guess we switch to prefetch then, however we can keep ascp for the ENA since it is actually much faster there.

Maarten-vd-Sande:
>download the requested file:
The file is downloaded using Aspera if available on your system, or HTTPS otherwise.

They can not be any more unclear?! I guess what they mean is that the alias ascp points to the path of ascp?

Maarten-vd-Sande:
No ascp with prefetch. Whatever, we can still download the fastqs directly from ena, and assume that if its not on ena as fastq then its not as sra either there..
https://github.com/ncbi/sra-tools/issues/255

Documentation is baaaad

Maarten-vd-Sande:
Prefetch makes is much, much more clean!

siebrenf (approved):
👍

Maarten-vd-Sande:
I downloaded 2 terabytes without problems!

Issue #482, opened 2020-08-28
BUG: conditions for macs2 peak calling


Maarten-vd-Sande:
**Describe the bug**
The current version does not work for narrowpeak_summit for macs2 and conditions. This is because Snakemake's ruleorder seems bugged?

We should split this rule into two version, one with macs2 as peak-caller, and the samples can only be the breps of multiple treps....

https://github.com/vanheeringen-lab/seq2science/blob/master/seq2science/rules/peak_counts.smk#L40-LL57

This is quite ugly, but I have no idea how else we can solve it

PR #483, opened 2020-08-28
Fix#482

Maarten-vd-Sande:
See issue #482. 

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 - update snakemake version
 
+### Fixed
+
+- bug with getting summit files from merged replicates of macs2
+
 ## [0.2.2] - 2020-08-24
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,6 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 ### Fixed
 
 - suffix being overwritten by layouts
-
-### Fixed
-
 - bug with getting summit files from merged replicates of macs2
 
 ## [0.2.2] - 2020-08-24
--- a/requirements.yaml
+++ b/requirements.yaml
@@ -5,7 +5,7 @@ channels:
   - defaults
 dependencies:
   - pkgs/main::python=3.6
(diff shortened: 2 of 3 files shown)


siebrenf (approved):

siebrenf:
I did some test with the old `ruleorder` back in (and no constraints) and it _seemed_ to work fine (again):

```
ruleorder: macs2_callpeak > narrowpeak_summit
rule narrowpeak_summit:
    """
    Convert a narrowpeak file to a "macs2 summits" file.
    """
    input:
        get_peakfile_for_summit,
    output:
        expand("{result_dir}/{{peak_caller}}/{{assembly}}-{{sample}}_summits.bed", **config),
    log:
        expand("{log_dir}/narrowpeak_summit/{{sample}}-{{assembly}}-{{peak_caller}}.log", **config),
    benchmark:
        expand("{benchmark_dir}/narrowpeak_summit/{{sample}}-{{assembly}}-{{peak_caller}}.benchmark.txt", **config)[0]
    shell:
        """
        awk 'BEGIN {{OFS="\t"}} {{ print $1,$2+$10,$2+$10+1,$4,$9; }}' {input} > {output} 2> {log}
        """
```

The docs comment that ruleorder does work if the input of a rule was missing. So perhaps this was due to the other bugs :s

Maarten-vd-Sande:
I thought before that it worked, however when I implemented it (already on master :scream: ) it failed.... But that still would be the best solution indeed!

https://github.com/vanheeringen-lab/seq2science/commit/4abd5070804dfe56a2a698c22b64f5f2379c5ecb

I can not find this in the docs:

> The docs comment that ruleorder does work if the input of a rule was missing. 

But that could indeed be the problem then.

p.s. tests are running 14 hours already :shrug:

siebrenf:
It's not listed with the wildcards, but with the [ambiguous rules sections](https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#snakefiles-ambiguous-rules). 

I have no clue about Jenkins. Have you tried turning it off and on again? :p

Maarten-vd-Sande:
I restarted my pc, lets see.

> e.g. due to missing input files

I think this does not mean inputerrors?

PR #493, opened 2020-09-03
remove dups

Maarten-vd-Sande:
I actually see no reason to keep duplicates. What's your opinion? @siebrenf 

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
You're right. We already measure duplication levels for the MultiQC, and multiple steps remove duplicates downstream as well. Plus, you made the change in the default config, so anyone can revert it if they want!

PR #494, opened 2020-09-03
fix bug merge replicates with condition

Maarten-vd-Sande:
When working with both control samples and treps, seq2science tries to merge_replicates out of thin air (no input, but output). This PR should solve that. I also removed the unnecessary moving and deleting files around when there are no multiple replicates in a trep.

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/merge_replicates.smk
+++ b/seq2science/rules/merge_replicates.smk
@@ -52,25 +52,16 @@ def rep_to_descriptive(rep, brep=False):
 if "replicate" in samples:
 
     def get_merge_replicates(wildcards):
-        input_files = dict()
-        input_files["reps"] = expand(
+        input_files = expand(
             [
                 f"{{trimmed_dir}}/{sample}{wildcards.fqext}_trimmed.{{fqsuffix}}.gz"
                 for sample in samples[samples["replicate"] == wildcards.replicate].index
             ],
             **config,
         )
-        if len(input_files["reps"]) == 0:
-            raise ValueError("Something went wrong, and we tried to merge replicates but there were no replicates?!")
-        # make sure we make the fastqc report before moving our file
-        if get_workflow() != "scatac_seq" and len(input_files["reps"]) == 1 and config["create_qc_report"]:
-            input_files["qc"] = expand(
-                [
-                    f"{{qc_dir}}/fastqc/{sample}{wildcards.fqext}_trimmed_fastqc.zip"
-                    for sample in samples[samples["replicate"] == wildcards.replicate].index
-                ],
-                **config,
-            )
+        if len(input_files) == 0:
+            return ["Something went wrong, and we tried to merge replicates but there were no replicates?!"]
+
         return input_files
 
     ruleorder: merge_replicates > renamefastq_PE > sra2fastq_PE
@@ -85,7 +76,7 @@ if "replicate" in samples:
         If a replicate has only 1 sample in it, simply move the file.
         """
         input:
-            unpack(get_merge_replicates),
+            get_merge_replicates,
         output:
             temp(sorted(expand("{trimmed_dir}/{{replicate}}{{fqext}}_trimmed.{fqsuffix}.gz", **config))),
         wildcard_constraints:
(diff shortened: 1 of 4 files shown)


seq2science/rules/merge_replicates.smk
```diff
@@ -83,9 +87,9 @@ if "replicate" in samples:
         input:
             unpack(get_merge_replicates),
         output:
-            temp(sorted(expand("{trimmed_dir}/merged/{{replicate}}{{fqext}}_trimmed.{fqsuffix}.gz", **config))),
+            temp(sorted(expand("{trimmed_dir}/{{replicate}}{{fqext}}_trimmed.{fqsuffix}.gz", **config))),
```
siebrenf, line 90: This is a BIG change. With this, we no longer know which samples are original, and which are replicates. Furthermore, replicate names can no longer be the same as sample names.

seq2science/rules/merge_replicates.smk
```diff
@@ -60,6 +60,8 @@ if "replicate" in samples:
             ],
             **config,
         )
+        if len(input_files["reps"]) == 0:
+            raise ValueError("Something went wrong, and we tried to merge replicates but there were no replicates?!")
```
siebrenf, line 64: Nice for debugging, but not very nice to leave in. In `renamefastq_PE` I solved this by adding an impossible input if the input was empty. This input can double as a debug message, if snakemake calls an `missinginputException` on it.

seq2science/rules/merge_replicates.smk
siebrenf: Nvm, only the first point stands, but I dont think thats important enough.

seq2science/rules/merge_replicates.smk
```diff
@@ -83,9 +87,9 @@ if "replicate" in samples:
         input:
             unpack(get_merge_replicates),
         output:
-            temp(sorted(expand("{trimmed_dir}/merged/{{replicate}}{{fqext}}_trimmed.{fqsuffix}.gz", **config))),
+            temp(sorted(expand("{trimmed_dir}/{{replicate}}{{fqext}}_trimmed.{fqsuffix}.gz", **config))),
         wildcard_constraints:
-            replicate=any_given("replicate", "control"),
+            replicate=any_given("replicate"),
```
siebrenf, line 92: what was going on here?

siebrenf (changes requested):
I was afraid of removing the merged folder, but it seems that might work out just fine! 
Just a couple of smaller comments left :)

seq2science/rules/merge_replicates.smk
Maarten-vd-Sande: Control samples needed to be downloaded and "merged" as well. However if we remove the /merged/ folder, then that is not necessary anymore.

seq2science/rules/merge_replicates.smk
```diff
@@ -71,6 +73,8 @@ if "replicate" in samples:
             )
         return input_files
 
+    ruleorder: merge_replicates > renamefastq_PE > sra2fastq_PE
```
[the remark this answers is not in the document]
Maarten-vd-Sande: I am not sure if we can remove it. If you only download fastqs, then merge_replicates does not exist, so you can not use that in the ruleorder. Plus the merge_replicates snakefile is not even loaded.

seq2science/rules/merge_replicates.smk
Maarten-vd-Sande: I think this is personal. I prefer to raise an error immediately than have snakemake try to make some file from some "impossible" output. It really should not happen! It's not like people put in the wrong sample name, this is a real problem if it happens.

Also using "impossible" output might result in weird bugs/side effects later, especially in combination with generic rules like unzip.

siebrenf: an impossible input prevents the rule from being used. no weird side effects. Something like
```
if len(input_files["reps"]) == 0:
    return "Something went wrong, and we tried to merge replicates but there were no replicates?!"
```

The reason I used this solution is because (I think) snakemake runs all these rules while trying to generate the DAG. So it might "try" the rule, then run into the error, even though it wouldn't use the rule in the end. Not sure if we test for this scenario.

siebrenf (approved):

Issue #445, opened 2020-08-03
FR: switch to pysradb (investigate feasibility) for checking layout


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
I dont want to wait for the lookup of sample layouts anymore!

**Describe the solution you'd like**
Seems like pysradb has the option to check for sample layout. Might be nice to switch :)
https://github.com/saketkc/pysradb
https://saket-choudhary.me/pysradb/python-api-usage.html#use-case-1-fetch-the-metadata-table-sra-runtable

No clue how well this works though.

Maarten-vd-Sande:
pysradb requires you to download a 25 gigabyte sqlite database before usage.....

Maybe not the best idea

saketkc:
Hi @Maarten-vd-Sande, Just to clarify, `pysradb` does not require `SRAmetadb.sqlite` downloads anymore (and hasn't required so for a while now). 

You can checkout the Colab notebooks for demonstration of the web based API:  https://github.com/saketkc/pysradb/tree/master/notebooks

P.S. I discovered this repo via Github. Seq2Science looks cool!

Maarten-vd-Sande:
Hi @saketkc, this is quite a coincidence!!! 

Just yesterday I discovered the `SRAweb` class and I have been working on implementing it in seq2science :tada:, since it works much nicer than the e-utils api. 

One question, but maybe I should make an issue about it, the `library_layout` seems to be missing. However in the docs they are [there](https://saket-choudhary.me/pysradb/usage_scenarios.html#scenario-1) when using the detailed flag. I am using version `0.11.0`. Also in the notebooks they seem to be missing.

PR #499, opened 2020-09-09
Use pysradb to do geoena lookups

### LINKED ISSUE #321: BUG: (public) samples mixed with single-end and paired-end data
**Describe the bug**
https://www.ncbi.nlm.nih.gov/sra/?term=GSM2156963 and https://www.ncbi.nlm.nih.gov/sra?term=SRX1424709 are a mix of single-end and paired-end runs. This now crashes.

**Expected behavior**
Not really sure. Ideally the pipeline either gracefully notifies the user of the issue, or somehow manages to combine these two datatypes.

Maarten-vd-Sande:
Use pysradb for lookups #445 . For GSM numbers we first convert them to SRX, and from the SRX we get the SRR numbers. Seems to take around 60 seconds for 300 samples, so slightly faster. However much more robust, since some samples are on GEO but not on SRA, so by converting to SRX we circumvent this problem.

**What did change**
A lot!

- when downloading a sample, it gets split into its runs. These runs are downloaded and dumped separately, and are merged together into a sample later 
- sample info is stored in sampledict variable
- added a util module
- sample layout comes from pysradb
- ena download links come from pysradb
- reverted back to parallel-fastq-dump since fasterq-dump somehow keeps placing intermediate files in cwd
- fix #321 
- unfixes #228 (do/should we care about these issues?)

**Checklist**
- [ ] I updated the CHANGELOG
- [ ] requires pysradb > 0.11.0


Maarten-vd-Sande:
@siebrenf I still have to test it with a buttload of data, but so far so good. Could you also try this with your RNA-seq data to see if this breaks anything and if it works as expected?

Maarten-vd-Sande:
Probably needs a `seq2science clean` before

seq2science/rules/configuration_generic.smk
```diff
@@ -332,259 +330,54 @@ else:
...
 logger.info("Done!\n\n")
-logger.info("Now checking if the samples are on the ENA database..")
-logger.info("This can also take some time!")
-
-ena_single_end_urls = dict()
-ena_paired_end_urls = dict()
 
-# trace_tp = ThreadPool(40)
+# now check where to download which sample
+ena_protocol = "ena_fastq_ftp" if config.get("ascp_path") and config.get("ascp_key") else "ena_fastq_http"
```
siebrenf, line 364: wget should work on both ftp and http, and speed is similar. This means we can set ftp as default, and fall back to http if it doesnt exist/work. This is not an immediate requirement tho.

requirements.yaml
```diff
@@ -19,3 +19,6 @@ dependencies:
   - conda-forge:pretty_html_table=0.9.dev0
   - conda-forge:pandas_schema=0.3.5
   - bioconda::trackhub=0.1.2019.12.24
+#  - bioconda::pysradb=0.11.0
+  - pip:
+      - git+https://github.com/saketkc/pysradb.git
```
siebrenf, line 24: norns and beautifulsoup4 (and perhaps others) are no longer required now right?

nitpick: newline

siebrenf (approved):

requirements.yaml
Maarten-vd-Sande: Yes, norns and bs4 you are correct! For the actual release we are waiting for pysradb 0.11.1, since that has some bug fixes we need, but is not released. So that will be replaced by the bioconda thing

PR #500, opened 2020-09-09
Update rulerorder rename_fastq

siebrenf:
minor fix


Maarten-vd-Sande (approved):

Maarten-vd-Sande:
We really should not be merging ruleorders before we update the snakemake version!! We have to be careful with these changes

PR #509, opened 2020-09-14
Explain fix

Maarten-vd-Sande:
small fixes to explain. Not sure how/if we should add this to the tests

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):

Maarten-vd-Sande:
you can't take it from the requirements file. It is outside of the seq2science folder, so we can't really ship it with the package. It also does not guarantee this version...

Maarten-vd-Sande:
Oeh if that works, then that's awesome!

PR #510, opened 2020-09-14
improved config & extended docs

siebrenf:
config could be clearer

**What did change**
- Several config options added to the schemas so they appear in the extended docs
- Renamed `spike_suffix` to `custom_assembly_suffix` as the feature applies to all additions. not just spike-ins.
- Split `diffexp` into `deseq2` and `contrasts` to improve readability.
- updated `explain()` on Markduplicates, and explain after running the configuration rules

**Checklist**
- [x] I updated the CHANGELOG


seq2science/rules/configuration_workflows.smk
```diff
@@ -184,6 +187,10 @@ def sieve_bam(configdict):
     )
 
 
+# after all is done, explain the workflow if requested
```
Maarten-vd-Sande, line 190: I think just as with importing files it's better to leave the includes on the top.

seq2science/schemas/config/alignment_general.schema.yaml
```diff
@@ -36,6 +36,12 @@ properties:
     enum: [fisher, idr, keep]
     type: string
 
+  # genomepy
+  provider:
+    description: specify to only download new genomes from one provider
```
Maarten-vd-Sande, line 41: Maybe add if unspecified tries to download from them in order of ...?

seq2science/schemas/config/gene_expression.schema.yaml
```diff
@@ -23,7 +23,7 @@ properties:
           quantifier_flags:
             default: '--seqBias --gcBias --validateMappings --recoverOrphans' # --posBias experimental flag
           decoy_aware_index:
-            default: False
+            default: False  # Requires ~65 GB RAM for GRCh38
```
Maarten-vd-Sande, line 26: The memory could be added as a resource?

Maarten-vd-Sande (approved):

seq2science/schemas/config/gene_expression.schema.yaml
siebrenf: It already is, but this is visible in the extended docs. Should I remove it?

seq2science/rules/configuration_workflows.smk
siebrenf: We're using includes in varying places already. However, this could be an opportunity to make the code a bit more pythonic. Ill make a 2nd PR with an idea...

siebrenf: see #511

PR #511, opened 2020-09-15
includes at the start of files ;)

siebrenf:
Includes are strewn about (configuration_workflows.smk and explain.smk has to be in specific places to work propperly).
Configs were getting large and convoluted.

**What did change**
configuration files split down further:
0: configuration.smk
1: configuration_general.smk
2: configuration_workflow.smk
3: configuration_explain.smk (renamed for consistency)
4: configuration_logging.smk

moved some functions to a util.py file, similar to PR #499.

**Note**
this is a PR to another PR. See if you think this makes sense, otherwise we cancel it.

**Next**
I wouldn't mind putting the layout tracing into a separate `configuration_sample_layout.smk`. 
We could move the config rules into a subdir too.

### DIFF
--- a/seq2science/rules/qc.smk
+++ b/seq2science/rules/qc.smk
@@ -1,5 +1,7 @@
 import re
 
+from seq2science.util import sieve_bam
+
 
 def samtools_stats_input(wildcards):
     if wildcards.directory == config["aligner"]:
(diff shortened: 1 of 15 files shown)


seq2science/rules/qc.smk
```diff
@@ -1,5 +1,7 @@
 import re
 
+from seq2science.util import sieve_bam
```
Maarten-vd-Sande, line 3: I'm on my phone, but I think you also need to import "just" seq2science here, because seq2science.__version__ is used in the qc report

Maarten-vd-Sande (approved):

Maarten-vd-Sande:
I like this a lot! It was starting to get really messy, and this cleans up that mess

PR #512, opened 2020-09-16
Fastp

Maarten-vd-Sande:
Added fastp (new default) as a trimmer in combination with trimgalore. Now you can choose!

It gives a nicer multiqc report since it also collapses for replicates: http://ocimum.science.ru.nl/multiqc_dm6.html. This can be implemented for trimgalore/fastqc as well of course, but I think that's outside of this PR. (get_trimming_qc needs some cleanup perhaps)

**Checklist**
- [x] I updated the CHANGELOG


seq2science/rules/qc.smk
```diff
@@ -555,21 +620,26 @@ rule multiqc:
...
-        return expand(f"{{qc_dir}}/fastqc/{sample}_{{fqext}}_trimmed_fastqc.zip", **config)
-    else:
-        if config['layout'][sample] == 'SINGLE':
-            return expand([f"{{qc_dir}}/fastqc/{sample}_fastqc.zip",
-                           f"{{qc_dir}}/fastqc/{sample}_trimmed_fastqc.zip",
-                           f"{{qc_dir}}/trimming/{sample}.{{fqsuffix}}.gz_trimming_report.txt"],
-                           **config)
+    if config["trimmer"] == "trimgalore":
+        if get_workflow() == "scatac_seq":
+            # we (at least for now) do not was fastqc for each single cell before and after trimming.
```
siebrenf, line 625: "we (at least for now) do not was fastqc"

so far this is the biggest issue I have been able to find :D

seq2science/rules/trimming.smk
```diff
@@ -1,76 +1,142 @@
...
+        output:
+            r1=temp(expand("{trimmed_dir}/{{sample}}_{fqext1}_trimmed.{fqsuffix}.gz", **config)),
+            r2=temp(expand("{trimmed_dir}/{{sample}}_{fqext2}_trimmed.{fqsuffix}.gz", **config)),
+            qc_json=expand("{qc_dir}/trimming/{{sample}}.fastp.json", **config),
+            qc_html=expand("{qc_dir}/trimming/{{sample}}.fastp.html", **config),
+        conda:
+            "../envs/fastp.yaml"
+        threads: 3
+        wildcard_constraints:
+            sample="|".join([sample if config["layout"][sample] == "PAIRED" else "$a" for sample in all_samples])
```
siebrenf, line 129: what is `$a` here? An impossible input, a warning message defined elsewhere?

seq2science/rules/peak_calling.smk
```diff
@@ -102,14 +102,16 @@ rule call_peak_genrich:
...
+def get_fastq_qc_file(wildcards):
+    if config["trimmer"] == "trimgalore":
+        if (
+            config["layout"].get(wildcards.sample, False) == "SINGLE"
+            or config["layout"].get(wildcards.assembly, False) == "SINGLE"
+        ):
+            return expand("{qc_dir}/fastqc/{{sample}}_trimmed_fastqc.zip", **config)
+        return sorted(expand("{qc_dir}/fastqc/{{sample}}_{fqext1}_trimmed_fastqc.zip", **config))
+    elif config["trimmer"] == "fastp":
+        return expand("{qc_dir}/trimming/{{sample}}.fastp.json", **config)
```
siebrenf, line 114: nice that macs2 can handle both!

siebrenf (approved):
the only (minor) gripe I could find was this vaguely named variable `$a`

seq2science/rules/trimming.smk
Maarten-vd-Sande: Yes, $ means end of string, so $a means match a letter a after the end of a string. Which can never be matched. We need to add those, because if we only add possible matches, if the list comprehension is empty we end up with an empty string which can always be matched.

seq2science/rules/qc.smk
Maarten-vd-Sande: I agree. Any proposals? The "problem" is that fastp does both qc and trimming, so not sure how to name it, but it is unclear indeed.

sidenote: get_trimming_qc needs cleanup, but I mentioned that before

seq2science/rules/trimming.smk
Maarten-vd-Sande: btw we need this (and not for fastqc) because ruleorder can not resolve only qc for technical replicates and no trimming

seq2science/rules/qc.smk
siebrenf: `# do not load all QC data for each single cell (its too much for MultiQC)`?

It doesn't clear up what we leave out, but the reason is stated.

Issue #515, opened 2020-09-18
BUG: error edgeR_normalization


dzijlmans:
**Describe the bug**
```
[Fri Sep 18 13:52:16 2020]
Error in rule edgeR_normalization:
    jobid: 284
    output: /home/dzijlmans/ChIP-seq/DZ023/results_resonicated_200917_complete/count_table/macs2/GRCm38.p6_upperquartile.tsv
    log: /home/dzijlmans/ChIP-seq/DZ023/results_resonicated_200917_complete/log/edgeR_normalization/GRCm38.p6-macs2-upperquartile.log (check log file(s) for error message)
    conda-env: /home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/site-packages/seq2science/.snakemake/ee92d238

RuleException:
CalledProcessError in line 249 of /home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/site-packages/seq2science/rules/peak_counts.smk:
Command 'source /home/dzijlmans/anaconda3/envs/seq2science2/bin/activate '/home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/site-packages/seq2science/.snakemake/ee92d238'; set -euo pipefail;  Rscript --vanilla /home/dzijlmans/ChIP-seq/DZ023/.snakemake/scripts/tmpi74ro7gu.edger_norm.R' returned non-zero exit status 1.
  File "/home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/site-packages/snakemake/executors/__init__.py", line 2135, in run_wrapper
  File "/home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/site-packages/seq2science/rules/peak_counts.smk", line 249, in __rule_edgeR_normalization
  File "/home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/site-packages/snakemake/executors/__init__.py", line 529, in _callback
  File "/home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/concurrent/futures/thread.py", line 57, in run
  File "/home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/site-packages/snakemake/executors/__init__.py", line 515, in cached_or_run
  File "/home/dzijlmans/anaconda3/envs/seq2science2/lib/python3.7/site-packages/snakemake/executors/__init__.py", line 2201, in run_wrapper

```

**To Reproduce**

*config.yaml*
```
[...]

PR #520, opened 2020-09-29
fix: 515, replace by all nan if edger fails

Maarten-vd-Sande:
See #515 

**What did change**
I added a trycatch, so when an error happens we make an array full of NA. @siebrenf I haven't fully tested this in snakemake so as the R expert, can you take a careful look :eyes: ?

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ### Fixed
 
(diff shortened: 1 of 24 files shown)


siebrenf (approved):
What is the added benefit of continuing after EdgeR fails? 

If `--keep-going` provides the same benefit, I would not do this, but if other analyses can still be performed, then sure 👍

Maarten-vd-Sande:
The problem is that s2s does a couple of different normalization methods. Somehow upperquartile manages to fail on a sample that exists of normal numbers, where TMM and quantile norm still work. I guess in this case its better to report a NA array, than worry the user with a failed rule?

Issue #471, opened 2020-08-18
FR: order of tracks the same as samples.tsv in trackhub


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
No clue how the order currently is decided, but would be nice if it follows this order. I nicely sorted my samples file, but every time I open the hub I have to slide all the tracks :(

PR #523, opened 2020-10-03
trackhub 2.0

siebrenf:
The tracks are now grouped, and can be filtered per replicate or per track type.
Also, colors are added automatically, or can be supplied from the samples.tsv

With the new grouping, all samples are controlled together, making scaling and editing much more easy.
  - @Rebecza mentioned how bigwig scaling can bedeceptive. So now, all bigWigs are scaled to the same y-axis by default.
    - and the axis is visible by default!

track view:
image

configuration options:
image

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
Maaaybe you can take a look at #471 :angel: 

I am not even sure whether or not the order is the same (or isnt)

Maarten-vd-Sande (approved):

siebrenf:
Before, we list all the files used in the trackhub in two places: once in the input function, and once in the rule.
With this commit, the files are aggregated into one dict. That dict can then be parsed by both the rule and the input function.

that should start making this behemoth a little more manageable...

Maarten-vd-Sande:
Makes sense!

Any opinion about https://github.com/vanheeringen-lab/seq2science/issues/527 ? I am not sure what would be the cleanest way to implement this..

siebrenf:
Next steps are adding color support and (trying to) add control groups.

For color support, here's my idea:
1) count the number of expected tracks (ATAC/ChIP: breps, RNA/alignment: treps)
2) create a colorspace over this number
3a) for ATAC/ChIP, add near-colors* for the treps within each brep
3b) for stranded RNA-seq, add near-colors for the reverse stranded track

Next, the user can add a column to the samples.tsv. If this column is present, this default colorset is overruled.
- empty fields are considered black
- for treps, the first populated field is used
- for breps... an average of the treps is used? with a near-color for the trep if there is only 1?

*near-color: a slighlty paler version or slightly brighter color.

what do you think? :)

Maarten-vd-Sande:
Yes I like this idea. 

Would be nice if we can use hex, RGB, and the default python color names (like 'b' for blue). There's probably a function built-in that does this, otherwise I can help with that. 

Another option would be to use numbers, and sample from the colorscale (e.g. think time-series). I can also help with that

Maarten-vd-Sande:
@JGASmits and @simonvh you both like colors. Anything to add?

simonvh:
For hex, RGB, and color name, this should be easily supported by matplotlib. For nice palettes, check the colorbrewer palettes (https://colorbrewer2.org/) also supported by matplotlib. Keep in mind that the lighter colors of most sequential color schemes are not very good for visibility in the genome browser. There's also [palettable](https://jiffyclub.github.io/palettable/).

I think choosing colors would go very well with some kind of grouping. For example, I would have ATAC-seq and two kinds of ChIP-seq, both as a time course. These would be three groups. For these three groups I would then choose a palette name (i.e. PuBuGn or PuRd, etc.). Then the tracks would automatically get the an X number of colors of the palette, with order same as the track order. 

However, it's easy to want to do to much straightaway. I would already be happy if I could specificy one color per track.

JGASmits:
Just to add my 2 cents, I think the option to give a hex code to your sample name is already amazing.
Regarding default colours of the trackhub I think it might get complicated and diverse really fast (depending in what people want to show/compare). But perhaps I'm a bit pessimistic there. 
So personally I would stick to a simple default colour scheme, and if people want fancy stuff just let them input their own HEX colours. But if you put in a really complecated default scheme that colours all my samples amazing, of course I also wouldnt mind.  ;)

JGASmits (approved):

siebrenf:
pandas doesnt like hexes to start with `#`. so adding colors by hex code isnt part of the package atm.

but look at these colors!
image

siebrenf:
btw, I started out with the rainbow color pattern, but i think it gets difficult to see where one sample starts and another ends: 
image

Maarten-vd-Sande:
Very cool!!

siebrenf:
Turns out composite tracks throw out all my previous refactoring -,-

But thats past now. ChIP-/ATAC-seq hub is pretty much finished, on to the Alignment/RNA-seq hubs!
image

siebrenf:
RNA-seq: check
tested for assembly hubs and trackhubs: check

image

seq2science/rules/configuration_generic.smk
```diff
@@ -201,6 +206,14 @@ def get_workflow():
     return workflow.snakefile.split('/')[-2]
 
 
+wf = get_workflow()\
+    .replace('alignment',  'Alignment')\
+    .replace('atac_seq',   'ATAC-seq')\
+    .replace('chip_seq',   'ChIP-seq')\
+    .replace('rna_seq',    'RNA-seq')\
+    .replace('scatac_seq', 'scATAC-seq')
```
Maarten-vd-Sande, line 214: Maybe we should rename `wf` to `workflow` or sth. Not sure how DeScRiPtIvE it is :P

Anyways, regardless whether or not you change the name, please add it to:
https://github.com/vanheeringen-lab/seq2science/blob/master/docs/CONTRIBUTING.md#important-global-variables

Maarten-vd-Sande (approved):
Looks good! :+1: 

One thing we could consider is moving the trackhub into a script (I believe something you proposed a year ago and I opposed :facepalm: ). It is getting bigger and bigger, plus a bonus side-effect is that the trackhub environment can be removed from the base environment.

But we could leave that for later as well


p.s. it looks soo much better with colors!

seq2science/rules/configuration_generic.smk
siebrenf: all valid points. 
 - wf renames to `sequencing_protocol` as workflow is taken by snakemake
 - I'll add it to the docs, but these aren't present in this PR, so later :)
 - I've quickly tried making this into a script, but there are a myriad of variables that this depends on (all_assemblies, has_annotation(), ori_assembly(), breps, treps, config...). It sounds like a nice challenge to set up. For another time ;)

siebrenf:
@Maarten-vd-Sande, any idea why snakemake.logging is duplicating everything?

Maarten-vd-Sande:
Nothing obvious... It does not happen in develop, or in the `download fastq` workflow. So it should be related to your changes :detective: 

The multiprocessing pool in `softmask...` is mildy suspicious, since it might just copy the complete environment (including logger). But I doubt that this one is causing any problems since it hasn't been changed in this branch.

The only other thing I can think of for now is perhaps matplotlib? I don't know if, maybe it does unexpected weird under-the-hood. You could try commenting it all out and not importing it, and see if that resolves the problem.

siebrenf:
the problem is importing trackhub. I need it in `create_trackhub()`, which is in the general environment.

importing (and deleting) the package in the function, or `import trackhub as th`, or `from trackhub import 1,2,3...` dont help. Guess I'll try that external script again...

siebrenf:
The issue was [here](https://daler.github.io/trackhub/_modules/trackhub/upload.html), specifically:
`logging.basicConfig(level=logging.INFO)`

I think it creates a basic everything-to-stdout setup out of the box. The fix now removes the last streamhandler from the list of streamhandlers. Since the handler has no name, this was the best I could come up with...

PR #531, opened 2020-10-10
does this resolve the FileNotFoundErrors?

Maarten-vd-Sande:
Sometimes the tests fail on the filelock:

```
FileNotFoundError in line 346 of /var/lib/jenkins/workspace/seq2science_develop/miniconda/envs/s2s/lib/python3.6/site-packages/seq2science/rules/configuration_generic.smk:
[Errno 2] No such file or directory: '/var/lib/jenkins/.config/seq2science/eutils.p.lock'
  File "/var/lib/jenkins/workspace/seq2science_develop/miniconda/envs/s2s/lib/python3.6/site-packages/seq2science/workflows/alignment/Snakefile", line 4, in <module>
  File "/var/lib/jenkins/workspace/seq2science_develop/miniconda/envs/s2s/lib/python3.6/site-packages/seq2science/rules/configuration.smk", line 1, in <module>
  File "/var/lib/jenkins/workspace/seq2science_develop/miniconda/envs/s2s/lib/python3.6/site-packages/seq2science/rules/configuration_generic.smk", line 346, in <module>
  File "/var/lib/jenkins/workspace/seq2science_develop/miniconda/envs/s2s/lib/python3.6/site-packages/filelock.py", line 323, in __enter__
  File "/var/lib/jenkins/workspace/seq2science_develop/miniconda/envs/s2s/lib/python3.6/site-packages/filelock.py", line 271, in acquire
  File "/var/lib/jenkins/workspace/seq2science_develop/miniconda/envs/s2s/lib/python3.6/site-packages/filelock.py", line 384, in _acquire
```

**What did change**
We make the lock after deleting it. This shouldn't really be necessary I think since we never made the lock beforehand I believe. But this should resolve the file not found error:

https://github.com/benediktschmitt/py-filelock/blob/master/filelock.py#L384

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 - bug with edgeR (upperquartile) normalization failed. Not sure why it fails, but when is does, it now returns a dataframe of nan instead of failing the rule, and thus the whole pipeline.
 - use gimmemotifs 0.15.0, so gimme.combine_peaks works with numeric chromosome names
+- touch a filelock after deleting it, to resolve filenotfounderrors 
 
 ## [0.3.0] - 2020-09-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 - bug with edgeR (upperquartile) normalization failed. Not sure why it fails, but when is does, it now returns a dataframe of nan instead of failing the rule, and thus the whole pipeline.
 - use gimmemotifs 0.15.0, so gimme.combine_peaks works with numeric chromosome names
-- touch a filelock after deleting it, to resolve filenotfounderrors 
+- s2s is slightly more lenient with an edge-case when running seq2science in parallel  
 
 ## [0.3.0] - 2020-09-22
 
--- a/seq2science/rules/configuration_generic.smk
+++ b/seq2science/rules/configuration_generic.smk
@@ -250,39 +250,44 @@ if "assembly" in samples:
     # determine provider for each new assembly
     providersfile = os.path.expanduser('~/.config/seq2science/providers.p')
     providersfile_lock = os.path.expanduser('~/.config/seq2science/providers.p.lock')
-    prep_filelock(providersfile_lock, 30)
-    with FileLock(providersfile_lock):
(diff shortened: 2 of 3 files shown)


siebrenf (changes requested):
If you create a brand new lock file, then the downstream `Filelock()` can only start if another process deletes the lock file. This can cause a hang on the last process.

I have a different suggestion using `pathlib`: 
replace `os.unlink(lock_file)` with `Path(lock_file).unlink(missing_ok=True)`. 
With this we can drop both `Path.touch()` and the `try except`!

Lemme know if Im missing something

siebrenf (commented):
Ok, so I've gone over this again, and I think neither of our solutions would fix the problem.

If I get it correctly, the issue is that the lock file is deleted by one process (using `prep_filelock`), WHILE it is expected by another process (using `FileLock`).

Perhaps the solution would be to use a retry wrapper around `FileLock`?
```
def filelock(lock_file, timeout):
    genomepy.utils.retry(func=FileLock, tries=2, lock_file, timeout)
```

Maarten-vd-Sande:
I am not sure what the problem is / causing this, but why would retrying work? The idea of deleting the lock is that maybe a seq2science instance exited incorrectly, leaving the lock on the file. 

Would the retry then solve the case when two instances are waiting for the lock at the same time?

siebrenf:
IIRC FileLock creates the lockfile if it is missing, or waits if it exists.

In your traceback, the error originates when `FileLock()` is called, but cannot find the lock file. So, my guess would be this following happened:
process 1: prep_filelock deletes and old lockfile. Next, FileLock creates a fresh lockfile,
process 2: prep_filelock deletes the lockfile new lockfile
process 1: hey wait!

the retry function has a 1 second delay, which should mean the 2nd process has passed beyond this point. If we have more than 2 processes running in parallel we just have to up the retry rate to match the number of processes.

Maarten-vd-Sande:
I think your explanation makes sense! I'll implement it. Thanks for the think-along

Maarten-vd-Sande:
I decided to put both the `prep_filelock` as the `with` statement in the "retry for loop".

siebrenf (approved):
what happens if it fails twice and hist the `continue`?
also, should there be a ~1sec wait in the `except`?

Maarten-vd-Sande:
Two very valid points! It should crash, and not just continue! And a sleep makes sense

PR #540, opened 2020-10-28
MultiQC many files

Maarten-vd-Sande:
For those people that do not know when to stop adding data... (I am looking at you @Rebecza !). I haven't tested it with a large amount of data, but if our assumption is correct that the max command length is overridden than MultiQC should now scale to an infinite amount of samples

**What did change**
Read the input files from a file, instead of from command line.

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
LOL. nice fix!

PR #542, opened 2020-10-28
stop trying when file exists locally

siebrenf:
When a GSM exists in the `fastq_dir`, `rule runs2sample` still tries to check `sampledict[wildcards.sample]["runs"]`. However, "runs" does not exist in this case.

**What did change**
prevented runs2sample from running its inputfunction when "runs" is not found in the sampledict.

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande (approved):

Maarten-vd-Sande:
Trying to wrap my head around why this isn't working :thinking:

Maarten-vd-Sande:
My guess is that `wildcard_constraints` actually are not evaluated on a "sample" basis, but on a rule basis. So once at the start of a workflow. I guess we could change it into an ugly list comprehension

PR #547, opened 2020-10-30
Update requirements.yaml

Maarten-vd-Sande:
A description of the problem that is being solved, or which new feature is implemented. Please link to other relevant issues/pr if relevant

**What did change**
Explain clearly and concisely what you changed. Why did you take certain design decisions?

**Checklist**
- [ ] I updated the CHANGELOG


Maarten-vd-Sande:
The problem here will be that somehow the newest snakemake version is not being approved on bioconda..

siebrenf:
installation succeeds locally with:

```
name: s2s
dependencies:
  - pkgs/main::python=3.8
  - bioconda::snakemake=5.27.4
  - bioconda::genomepy=0.9.1
  - bioconda::pysam=0.16.01
  - pkgs/main::conda=4.8.4
  - anaconda::biopython=1.74
  - pkgs/main::filelock=3.0.12
  - pkgs/main::pyyaml=5.3.1
  - conda-forge::pretty_html_table=0.9.dev0
  - conda-forge::pandas_schema=0.3.5
  - bioconda::trackhub=0.1.2019.12.24
  - bioconda::pysradb=0.11.1
  - conda-forge::matplotlib=3.3.2

```
pysam was updated (still no channels needed)

siebrenf:
```
[Tue Nov  3 20:21:15 2020]
rule get_genome:
    ....

[Tue Nov  3 20:21:15 2020]
rule get_genome_annotation:
    ....

Missing input files for rule get_genome_annotation:
/var/lib/jenkins/workspace/seq2science_PR-547/tests/local_test_results/fr3/fr3.fa
```

similar to `get_genome_support_files`, this rule does not seem to wait for the input rule to finish. Perhaps `run` is to blame? 
Either way, the version of genomepy does not matter.

Maarten-vd-Sande:
I am planning to release a new version soon (tomorrow?). Maybe we shouldn't update the snakemake version anyways that soon before a release, since last time was a disaster :P

siebrenf:
agreed, but lets upgrade genomepy! [Ive made a PR just for this](https://github.com/vanheeringen-lab/seq2science/pull/553)

Maarten-vd-Sande:
Ok so first of all `ancient` does not seem to play well with our tests in the new snakemake version. Also somehow all aligners seemed to fail (except star)...

What is happening between versions?!

siebrenf:
lets check with a smaller difference(?)

siebrenf:
it seems snakemake no longer makes the output path before the rule execution. The following commit should fix the rules (and updates some more versions that work locally)

Maarten-vd-Sande:
(to me) it's really strange they changed this behaviour. Does not seem to have any mention in the changelog.

siebrenf:
it gets weirder...

the current bug is caused by `configuration_generic` rerunning for each rule when using
`seq2science run atac-seq --cores 28 --configfile tests/alignment/remote_genome_n_sample.yaml`.

each time it does, `config["create_qc_report"]` is turned to string.

I have no idea why that smk is rerun, and why ONLY that smk is!

Maarten-vd-Sande:
Yes I saw that.. Not sure. I thought we maybe used the wrong key in the yaml, but that does not seem to be the case. All our stuff is rerun for each rule always (also why we need to cache stuff, otherwise we do way too many pysradb lookups). But indeed very strange that it works for some, but not rules. And that it crashes halfway... Not sure what to think of this.

siebrenf:
Do you know why the workflow reruns for each rule? It might point to the 2nd mystery :o

Maarten-vd-Sande:
I think that is just the way snakemake works... For each rule it starts a snakemake subprocess or sth, which needs to parse the samples and config file again. Not sure why there would be a difference in how it parses the config though... :thinking:

Maarten-vd-Sande:
image

siebrenf:
I think something goes wrong because after the first run, we edited the config. We never accounted for the possibility that the parsed config would be fed into the script again.

One option would be to rename config to parsed_config, so the original config remains unchanged. Ugly, but clean.

Another option is to use caches to circumvent the problem. The following solution works(!), but we'd need to use a thread-safe cache, and expand this for all configuration rules...

```
# config.yaml(s)

from appdirs import user_cache_dir
from bucketcache import Bucket

my_cache_dir = os.path.join(user_cache_dir("s2s_test"), seq2science.__version__)
if not os.path.exists(my_cache_dir):
    os.makedirs(my_cache_dir)
cache = Bucket(my_cache_dir, minutes=2)

@cache(ignore=["config"])
def config_gen(config, a_hash_to_recognize_this_run_by):
    # check config file for correct directory names
    for key, value in config.items():
        if '_dir' in key:
            assert value not in [None, '', ' '], f"\n{key} cannot be empty. For current directory, set '{key}: .'\n"
            # allow tilde as the first character, \w = all letters and numbers
            # ignore on Jenkins. it puts @ in the path
            assert (re.match('^[~\w_./-]*$', value[0]) and re.match('^[\w_./-]*$', value[1:])) \
                   or os.getcwd().startswith('/var/lib/jenkins'), \
                (f"\nIn the config.yaml you set '{key}' to '{value}'. " +
                  "Please use file paths that only contain letters, " +
                  "numbers and any of the following symbols: underscores (_), periods (.), " +
                  "and minuses (-). The first character may also be a tilde (~).\n")
            config[key] = os.path.expanduser(value)

    # make sure that difficult data-types (yaml objects) are in correct data format
    for kw in ['aligner', 'quantifier', 'bam_sorter', "trimmer"]:
        if isinstance(config.get(kw, None), str):
            config[kw] = {config[kw]: {}}

    # validate and complement the config dict
    for schema in config_schemas:
        validate(config, schema=f"{config['rule_dir']}/../schemas/config/{schema}.schema.yaml")


    # check if paired-end filename suffixes are lexicographically ordered
    config['fqext'] = [config['fqext1'], config['fqext2']]
    assert sorted(config['fqext'])[0] == config['fqext1'], \
        ("\nThe paired-end filename suffixes must be lexicographically ordered!\n" +
         f"Example suffixes: fqext1: R1, fqext2: R2\n" +
         f"Your suffixes:    fqext1: {config['fqext1']}, fqext2: {config['fqext2']}\n")

    # read the config.yaml (not the profile)
    with open(workflow.overwrite_configfiles[0], 'r') as stream:
        user_config = yaml.safe_load(stream)

    # make absolute paths, nest default dirs in result_dir and cut off trailing slashes
    config['result_dir'] = re.split("\/$", os.path.abspath(config['result_dir']))[0]

    if not url_is_alive(config['samples']):
        config['samples'] = os.path.abspath(config['samples'])

    for key, value in config.items():
        if key.endswith("_dir"):
            if key in ['result_dir', 'genome_dir', 'rule_dir'] or key in user_config:
                value = os.path.abspath(value)
            else:
                value = os.path.abspath(os.path.join(config['result_dir'], value))
            config[key] = re.split("\/$", value)[0]

    return config


@config_gen.callback
def config_gen(callinfo):
    print("\nCache used.\n")

a_hash_to_recognize_this_run_by = 1
config = config_gen(config, a_hash_to_recognize_this_run_by)


# samples.tsv
```

Maarten-vd-Sande:
The error seems to be related to:
https://github.com/snakemake/snakemake/pull/709

Maarten-vd-Sande:
No clue why packages do not want to build...

Maarten-vd-Sande:
I think my PR fixing the True thingy for snakemake is accepted and released
:pray:

Maarten-vd-Sande:
Ancient still dumb...

Maarten-vd-Sande:
The test also fails for me:

```
# variables used for this analysis:
salmonindex <- "/home/sande/Desktop/seq2science/tests/tinydata_custom/index/salmon_decoy_aware"
source      <- "Ensembl"
organism    <- "tximeta_file"
release     <- 42
genome      <- "tinydata_custom"
fasta       <- "/home/sande/Desktop/seq2science/tests/tinydata_custom/tinydata_custom.transcripts.fa"
gtf         <- "/home/sande/Desktop/seq2science/tests/tinydata_custom/tinydata_custom.annotation.gtf"
output      <- "/home/sande/Desktop/seq2science/tests/tinydata_custom/index/tximeta/linked_txome.json"
log_file    <- "/home/sande/Desktop/seq2science/tests/local_test_results/log/counts_matrix/tinydata_custom-linked_txome.log"

Sessioninfo:
R version 3.6.1 (2019-07-05)
Platform: x86_64-conda_cos6-linux-gnu (64-bit)
Running under: Ubuntu 18.04.5 LTS

Matrix products: default
BLAS/LAPACK: /home/sande/anaconda3/lib/R/lib/libRblas.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=nl_NL.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=nl_NL.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=nl_NL.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=nl_NL.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_3.6.1

Renaming GTF:
Error in loadNamespace(name) : there is no package called ‘R.utils’
Calls: :: ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
```

siebrenf:
I still don't know why we seemed to have some of the code from develop running here.

But I did finally figured out the bug:
the new version of tximeta apparently cant work with the test genome. My local PC had a cached txome from the previous version. Once I cleared the cache I could debug the problem. Tximeta now uses `source: "de-novo"` which fixes the issue.

siebrenf:
I think the error in the last commit was linked to the unknown behavior we saw. 

It might have something to do with the way Jenkins pulls the data. Over at http://ocimum.science.ru.nl/jenkins/job/seq2science/configure, the setting for "Discover pull requests from origin" was "Merging the pull request with the current target branch revision". I set it to "The current pull request revision" now.

Maarten-vd-Sande:
> I think the error in the last commit was linked to the unknown behavior we saw.
> 
> It might have something to do with the way Jenkins pulls the data. Over at http://ocimum.science.ru.nl/jenkins/job/seq2science/configure, the setting for "Discover pull requests from origin" was "Merging the pull request with the current target branch revision". I set it to "The current pull request revision" now.

from the naming it makes sense.



> I still don't know why we seemed to have some of the code from develop running here.
> 
> But I did finally figured out the bug:
> the new version of tximeta apparently cant work with the test genome. My local PC had a cached txome from the previous version. Once I cleared the cache I could debug the problem. Tximeta now uses `source: "de-novo"` which fixes the issue.

Awesome, now we have to decide what we do with `ancient` (And I will clean up all the True/true stuff)

requirements.yaml
```diff
@@ -4,10 +4,11 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - pkgs/main::python=3.6
-  - bioconda::snakemake=5.18.1
+  - pkgs/main::python=3.8
+  - bioconda::snakemake=5.30.1
+  - pkgs/main::conda=4.9.2
+  - conda-forge::mamba=0.7.1
```
Maarten-vd-Sande, line 10: are we going to add mamba as requirement? Might make sense?

seq2science/rules/alignment.smk
```diff
@@ -29,6 +29,8 @@ if config["aligner"] == "bowtie2":
             config["index"],
         shell:
             """
+            mkdir -p {output}
```
Maarten-vd-Sande, line 32: Are these necessary? The new snakemake version does not make dirs anymore?

Maarten-vd-Sande:
I kinda prefer `true` over `True` actually

requirements.yaml
siebrenf: why not? we run all our tests with mamba, so it is guaranteed to work!

seq2science/rules/alignment.smk
siebrenf: When I added them it was indeed required. 
You could remove them to see if the bugfix releases changed that.

siebrenf:
Then we keep `true` :)
and I guess we lose the `ancient()` if it's still broken (but lets add a #TODO).

siebrenf:
Why is the RNA-seq step is stuck on rule `softmask_track_1` for over 22h?
Could parallel runs cause interference with multiprocess Pool, or is it snakemake?

the rule:
https://github.com/vanheeringen-lab/seq2science/blob/b872cb34e53cdffe7016a6b1d852c64eae24a8b6/seq2science/rules/trackhub.smk#L137

Maarten-vd-Sande:
I don't know, that's weird! I restarted the test.. But that is suspicious! :thinking: 

I shouldn't really matter I guess

seq2science/envs/tximeta.yaml
```diff
@@ -6,11 +6,8 @@ channels:
...
-  # linked_txome
-  - r::r-r.utils=2.8.0
+  - conda-forge::r-base=4.0.3
+  - bioconda::bioconductor-tximeta=1.8.0
   # txi counts table
-  - bioconda::bioconductor-singlecellexperiment=1.8.0
-  - r::r-tidyverse=1.2.1
-  - r::r-readr=1.3.1
+  - bioconda::bioconductor-singlecellexperiment=1.12.0
+  - conda-forge::r-tidyverse=1.3.0
```
Maarten-vd-Sande, line 13: @siebrenf this stuff was done in the previous release right?

seq2science/schemas/config/gene_expression.schema.yaml
```diff
@@ -45,16 +45,20 @@ properties:
...
-        default: tximeta_file # organism (e.g. "Homo sapiens")
+        description: organism (e.g. "Homo sapiens")
+        default: tximeta_file
       release:
-        default: 42 # release number (e.g. "27")
+        description: release number of the source database (e.g. "101")
+        default: 42
     default:
-      source: Ensembl
+      source: de-novo
```
Maarten-vd-Sande, line 61: @siebrenf same here?

seq2science/scripts/generate_linked_txome.R
```diff
@@ -40,17 +35,20 @@ cat('\n')
 ## create a symlink to the gtf with Ensembl naming scheme (required for tximeta)
 fake_gtf_path <- file.path(dirname(output), paste0(organism, '.', genome, '.', release, '.gtf'))
```
Maarten-vd-Sande, line 36: and in here?

Maarten-vd-Sande (commented):
All those salmon and tximeta stuff is not relevant for this PR right? That was merged in previous version?

seq2science/scripts/generate_linked_txome.R
siebrenf: I ran into issues only during the tests in this PR

seq2science/schemas/config/gene_expression.schema.yaml
siebrenf: See below. The fix was to use the new "de-novo" option in this version of tximeta. lets see if it still works!

seq2science/scripts/generate_linked_txome.R
siebrenf: as there were merge conflicts relating to tximeta, i guess the main branch caught up with whatever was causing them. I've combined the fixes (I think).

siebrenf:
not sure if I'm overtaxing your PC, or if this is the new and "improved" DAG generation from snakemake, but the DAG tests take a lot longer!

siebrenf:
all core packages in requirements.yaml have been updated, `ancient()` is back in, and the tximeta stuff was merged. General tests seem to work, but I'll re-run an extensive RNA-seq analysis to be sure!

Maarten-vd-Sande:
> not sure if I'm overtaxing your PC, or if this is the new and "improved" DAG generation from snakemake, but the DAG tests take a lot longer!

yeah strange :thinking: 

I have been moving lots of files around tonight so maybe it was somehow IO related? The changelog said it would be faster

PR #554, opened 2020-11-04
strandedness fixes

siebrenf:
RNA-seq bigwigs only showed the reverse strand. oops.

**What did change**
RNA-seq bigwigs now show both strands. With this change, all references to strandedness can be dropped from non-RNA-seq workflows!

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/bigfiles.smk
+++ b/seq2science/rules/bigfiles.smk
@@ -134,6 +134,14 @@ rule peak_bigpeak:
         """
 
 
+def set_strand(wildcards):
+    if wildcards.strand == ".fwd":
+        return "--filterRNAstrand forward"
+    elif wildcards.strand == ".rev":
+        return "--filterRNAstrand reverse"
+    return ""
+
+
 rule bam_bigwig:
     """
     Convert a bam file into a bigwig file.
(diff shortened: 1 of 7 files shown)


Maarten-vd-Sande (approved):
can't say I fully understand whats going on here :angel:

siebrenf:
before the trackhub rework I tried to be too fancy: If the sequencing protocol was reverse-stranded, then I wanted the strands to be inverted in the bigwig. I think those two are unrelated however, so this change removes it (and fixes 2 serious trackhub bugs).

So what happens in the first commit: 
- Bigwig strand filtering is now handled in bigfiles.smk
- Trackhub bigwig selection is now handled in trackhub.smk
- With strandedness and bigwigs uncoupled, strandedness.smk is only needed for RNA-seq.
    - the smk is no longer imported anywhere except in RNA-seq
    - inside strandedness.smk, the functions no longer need to be inside an `if get_workflow() == "rna_seq"`

Maarten-vd-Sande:
Ok clear, can this be merged? Then I will retry our release. The previous trackhub bug was causing issues in the conda tests

PR #555, opened 2020-11-04
Maarten vd sande patch 1

Maarten-vd-Sande:
lists are mutable, so this just made subset and cols point to the same list. 

**Checklist**
- [ ] I updated the CHANGELOG


siebrenf (approved):
Good catch, and thanks for explaining!

Issue #578, opened 2020-11-19
FR: check size or shasum of downloaded fastq


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
Currently ENA is down and empty files are being sent. We should check if the file we get is not empty, and/or that the shasum corresponds.

Maarten-vd-Sande:
I can't seem to find the md5 files for fastqs on ENA..

PR #579, opened 2020-11-19
Emptyfile

Maarten-vd-Sande:
See issue #578 . Not sure why and what is happening at ENA, but now we at least check if the downloaded file contains at least 1 byte of info :joy: . I don't see md5 files for the fastq stuff on ENA, so that would be difficult to implement, so this is I guess the best we can do for now? Files from SRA do have md5 stuff I believe, but 90%+ is downloaded directly from ENA, so that's a minor improvement.

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/get_fastq.smk
+++ b/seq2science/rules/get_fastq.smk
@@ -193,8 +193,8 @@ rule ena2fastq_SE:
             shell("wget {url} -O {output} --waitretry 20 >> {log} 2>&1")
 
         if not (os.path.exists(output[0]) and os.path.getsize(output[0]) > 0):
-            shell("echo \"Something went wrong.. The downloaded file was empty!\\n\" >> {log} 2>&1")
-            shell("exit 1 >> {log} 2>&1")
+            shell('echo \"Something went wrong.. The downloaded file was empty!\" >> {log} 2>&1')
+            shell('exit 1 >> {log} 2>&1')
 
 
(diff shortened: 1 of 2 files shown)


siebrenf (changes requested):
Wont this only give the error if the file does not exist and is bigger than 0?

Or size == 0

seq2science/rules/get_fastq.smk
```diff
@@ -217,6 +221,11 @@ rule ena2fastq_PE:
             shell("wget {urls[0]} -O {output[0]} --waitretry 20 >> {log} 2>&1")
             shell("wget {urls[1]} -O {output[1]} --waitretry 20 >> {log} 2>&1")
 
+        if not (os.path.exists(output[0]) and (os.path.getsize(output[0]) > 0) and
```
siebrenf, line 224: If one of the pairs is 0, you can give an error too IMO

Maarten-vd-Sande:
:thinking: 

The way I see it is that I check for 2 conditions (or four if PE). Since I check the conditions with `and`, both have to be true for it to be true. Then I take the "opposite" of it, so that means to accept the if statement, at least one condition has to be false.

But I can simplify the logic by checking for `== 0`

siebrenf:
Ah! I think you need to use `not (a and b)` then?

siebrenf (approved):
You are. Me blind

PR #581, opened 2020-11-19
support rerunning

siebrenf:
**What did change**
added --rerun-updated CLI flag

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+### Added
+
+- --rerun-updated flag to the seq2science CLI
+
 ### Fixed
 
 - make a trackhub index when the gene_name is not present in gtf file
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -13,10 +13,13 @@ user@comp:~$ conda config --set channel_priority flexible
 ```
 
 ## I changed the config/samples file but seq2science does not rerun
-TODO: we should support rerunning!
-Seq2science (actually Snakemake) has a "lazy" policy regarding the generation of files, and will normally only rerun jobs if the input is younger than the output. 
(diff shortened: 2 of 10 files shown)


siebrenf:
from a quick check, I guess that `--list-params-changes` shows all accessed files, not just changed files.
with this PR, we take the input to the snakefile's final rule (renamed seq2science for ego and convenience), and force-rerun this.

it doesnt capture all changes, but almost all.

Maarten-vd-Sande:
I just took a quick glance, and based on your comment, I don't think this is the way forward. If we want to support rerunning, we should do it properly. So we should add all the used params as params (and samples!) to the rules. This is kinda a lot of work and error prone, but I don't mind spending a day on it next week. 

I haven't tried if that would work, but then we support full rerunning, instead of some in-between rerunning. What do you think?

siebrenf:
I agree full-support would be nicer. This is what we would need to do:
- per rule executed, store the input and params **upon rule completion** (to ensure the user can try mutliple things in dryrun mode before actually rerunning)
- when adding the rerun flag, check which input and params would change **per rule** (force rerun all in dryrun mode i guess)
- add the rules to the force rerun list

If you think that is worth implementing, sure! It would take a while though, and I'm not sure its worth the effort. If you think you can do this within a month or so, we can forget about this PR. If it would take longer, would it be worth implementing this for now? 

Either way, no hurt feelings if you think this approach is too bad to pull!

Maarten-vd-Sande:
I would have to check, but my guess would be that `--list-params-changes` does not store the dryrun stuff?

> List all output files for which the defined params have changed in the Snakefile.

My idea would be to delete the files with changed params, ask snakemake which rules it needs to generate those, and then force with those rules? 

It is still a lot of work of adding everything to every rule though, but then we do not need to store this stuff ourselves.

Maarten-vd-Sande:
But I  havent tried any of this stuff, so there might be an important part missing / not supported and that I am not aware of

Maarten-vd-Sande:
Closing this for now

PR #585, opened 2020-11-24
Rerun

Maarten-vd-Sande:
Finally had the time to work on this. Changing parameters should now result in rerunning :tada: :fire: 

One thing we need to **double check** is whether the relevant params have been added to each rule. Otherwise it gives a fake sense of security :confused: 

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
Another issue is keeping the stdout nice... :thinking:

Maarten-vd-Sande:
Ah, I was running this on snakemake version 5.27 on my own computer. Seems to work on that version though :upside_down_face:

seq2science/cli.py
```diff
@@ -287,8 +293,51 @@ def _run(args, base_dir, workflows_dir, config_path):
...
+            parsed_args["keep_logger"] = True
+
+    logger.info("Done. Now starting the real run.")
+    logger.stream_handler.setStream(sys.stdout)
+    parsed_args["config"]["no_config_log"] = True
+    #   5. start the "real" run where jobs actually get started
     exit_code = snakemake.snakemake(**parsed_args)
+
+    #   6. start the "real" run where jobs actually get started
     sys.exit(0) if exit_code else sys.exit(1)
```
siebrenf, line 341: same comment

seq2science/rules/get_genome.smk
```diff
@@ -32,7 +32,7 @@ rule get_genome_annotation:
     Download a gene annotation through genomepy.
     """
     input:
-        ancient(rules.get_genome.output),
+        rules.get_genome.output,
```
siebrenf, line 35: still causing issues? it seemed to work in my trial runs...

seq2science/util.py
```diff
@@ -407,3 +408,132 @@ def shorten(string, max_length, methods="right"):
         string = string[:ceil(max_length/2)] + string[len(string)-floor(max_length/2):]
 
     return string
+
+
+class CaptureStdout(list):
+    """
+    Context manager that somehow manages to capture prints,
+    and not snakemake log
```
siebrenf, line 416: 🙊

seq2science/util.py
```diff
@@ -407,3 +408,132 @@ def shorten(string, max_length, methods="right"):
...
+        samples['condition'] = samples['condition'].mask(pd.isnull, samples['replicate']) if 'replicate' in samples else \
+            samples['condition'].mask(pd.isnull, samples['sample'])
+        if samples['condition'].tolist() == samples['sample'].tolist() or config.get('biological_replicates') == 'keep':
+            samples = samples.drop(columns=['condition'])
+    if 'descriptive_name' in samples:
+        samples['descriptive_name'] = samples['descriptive_name'].mask(pd.isnull, samples['replicate']) if \
+            'replicate' in samples else samples['descriptive_name'].mask(pd.isnull, samples['sample'])
+        if ('replicate' in samples and samples['descriptive_name'].to_list() == samples['replicate'].to_list()) or \
+                samples['descriptive_name'].to_list() == samples['sample'].to_list():
+            samples = samples.drop(columns=['descriptive_name'])
```
siebrenf, line 497: these two were updated.

seq2science/util.py
```diff
@@ -407,3 +408,132 @@ def shorten(string, max_length, methods="right"):
...
+        sys.stderr = self._stringio = StringIO()
+        return self
+
+    def __exit__(self, *args):
+        self.extend(self._stringio.getvalue().splitlines())
+        del self._stringio  # free up some memory
+        sys.stderr = self._stderr
+
+
+def parse_samples(config):
```
siebrenf, line 445: this whole section is still present in config_generic.smk, and the function is unused(?)

seq2science/util.py
```diff
@@ -407,3 +408,132 @@ def shorten(string, max_length, methods="right"):
...
+                                     f"Condition {cond} can not also occur as a replicate"
+
+    # validate samples file
+    for schema in sample_schemas:
+        validate(samples, schema=f"{config['rule_dir']}/../schemas/samples/{schema}.schema.yaml")
+
+    sanitized_samples = copy.copy(samples)
+
+    samples = samples.set_index('sample')
+    samples.index = samples.index.map(str)
```
siebrenf, line 539: if this becomes an external function, shouldn't `samples` be returned, and `sanitized_samples` be copied outside this function?

On a different note: do we need a separate `sanitized_samples`?

seq2science/rules/bam_cleaning.smk
```diff
@@ -24,6 +24,9 @@ rule setup_blacklist:
         unpack(get_blacklist_files),
     output:
         temp(expand("{genome_dir}/{{assembly}}/{{assembly}}.customblacklist.bed", **config)),
+    params:
+        config.get("remove_blacklist"),
+        config.get("remove_mito"),
```
siebrenf, line 29: add `  # helps resolve changed params`

seq2science/rules/get_fastq.smk
```diff
@@ -60,7 +60,7 @@ rule sra2fastq_SE:
     benchmark:
         expand("{benchmark_dir}/sra2fastq_SE/{{run}}.benchmark.txt", **config)[0]
     wildcard_constraints:
-        run="SRR\d+",
+        run="|".join(sra_single_end) if len(sra_single_end) else "$a",
```
siebrenf, line 63: this is to prevent snakemake from accusing this rule right? a comment would be nice.

seq2science/rules/get_fastq.smk
```diff
@@ -102,7 +102,7 @@ rule sra2fastq_PE:
         expand("{benchmark_dir}/sra2fastq_PE/{{run}}.benchmark.txt", **config)[0]
     threads: 8
     wildcard_constraints:
-        run="SRR\d+",
+        run="|".join(sra_single_end) if len(sra_paired_end) else "$a",
```
siebrenf, line 105: same as above

seq2science/rules/qc.smk
```diff
@@ -503,6 +505,8 @@ rule multiqc_filter_buttons:
     """
     output:
         temp(expand('{qc_dir}/sample_filters_{{assembly}}.tsv', **config))
+    params:
+        samples
```
siebrenf, line 509: add `  # helps resolve changed params if e.g. descriptive names change`

seq2science/rules/qc.smk
```diff
@@ -518,7 +522,8 @@ rule multiqc_samplesconfig:
     params:
         config_used=len(workflow.overwrite_configfiles) > 0,
         configfile=workflow.overwrite_configfiles[-1],
-        sanitized_samples=sanitized_samples
+        sanitized_samples=sanitized_samples,
+        config={k: v for k, v in config.items() if k != "no_config_log"}
```
siebrenf, line 526: add `  # helps resolve changed config options`

siebrenf (changes requested):
This is a good solution! 

I personally dislike the params that don't affect the rule (it feels like this should be in snakemake itself), but it is a good solution. Most remarks are missing comments for these extra params.

`utils.parse_samples()` is outdated and unused, correct?

seq2science/rules/get_genome.smk
Maarten-vd-Sande: Nope, I guess that stayed from the merge from the update_snakemake branch from before, sloppy

seq2science/util.py
Maarten-vd-Sande: :shrug:

seq2science/util.py
Maarten-vd-Sande: good catch, The merge just auto-merged this stuff :eyes:

seq2science/util.py
Maarten-vd-Sande: !

Maarten-vd-Sande (commented):
Thanks for the review. The auto-merge really just skipped a big portion!

seq2science/cli.py
```diff
@@ -287,8 +293,51 @@ def _run(args, base_dir, workflows_dir, config_path):
     parsed_args["config"].update({"cores": parsed_args["cores"]})
     resource_parser(parsed_args)
 
-    # run snakemake
+    # run snakemake/seq2science
+    #   1. pretty welcome message
+    setup_logger()
+    log_welcome(logger, parsed_args)
+    if not args.skip_rerun or args.unlock:
```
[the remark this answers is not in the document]
siebrenf: it would be nice if `--skip-rerun` is on when you haven't started a run yet. 
This could be accomplished with `if os.path.exists(config["results_dir"]) and (not args.skip_rerun or args.unlock):`

buuuut this may be too tedious for a one-time benefit! I just realized we dont have access to `config` yet.

seq2science/util.py
Maarten-vd-Sande: why does this not make some tests fail?! :warning:

Maarten-vd-Sande: not used... :facepalm:

seq2science/util.py
Maarten-vd-Sande: Not sure if we need a separate one.. It's not really used

siebrenf (approved):

Maarten-vd-Sande:
@siebrenf if users now add samples to the RNA-seq workflow this works :tada: . However one "problem" now is that the strandedness is inferred from an intermediate bam file, not from the final bam. This means that if you add samples you have to re-do almost all steps. Perhaps I should change the strandedness to be done on the final bam? What do you think? Probably a fairer/better estimation anyways, since duplicates e.g. are removed (and don't get counted multiple times like that).

siebrenf:
that should work!

Maarten-vd-Sande:
@siebrenf I think I know what was the problem when adding more samples. The blacklist was recreated, which resulted in all mapping steps to be redone. By keeping the blacklist (with the blacklist config as params), this is now solved. However another problem remains for RNA-seq, since that puts all the samples together in a strandedness report. When adding new samples, the strandedness report is remade, and all downstream rules re-run. This means the quantification for all samples. It's not the end of the world, but I will make an issue about it.

PR #591, opened 2020-11-27
fix #587

### LINKED ISSUE #587: BUG: duplicate rows in count table
**Describe the bug**
```
Warning message:
package ‘edgeR’ was built under R version 3.6.2 
```

**To Reproduce**
I ran the atac-seq pipeline with default settings

Maarten-vd-Sande:
Fixes an edge case where taking the slop of peaks at the edge of a chromosome results in identical peaks, making edgeR unhappy. This barely ever happens, and the one case it did happen for me it was two peaks in 62000 peaks. :detective: 

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


siebrenf (approved):
`uniq` only checks the last line. should be fine as long as the output is already sorted.

Maarten-vd-Sande:
I did not know that :point_up_2: . Seemed to work in this case, so I am assuming its sorted

PR #595, opened 2020-12-01
fix trackhub naming bug

Maarten-vd-Sande:
Small bug in trackhub names for atac and chip. This caused collisions and the whole trackhub did not load

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


siebrenf (approved):
didn't know that could cause an error! I though the suffix (bw/pk) was enough to keep them unique?

The changes to `rep_to_descriptive` make sense either way!

Maarten-vd-Sande:
So there were two things and I didn't explain it properly. The peak would be called e.g. `24hpf_rep1 pk`, from the descriptive_name, but that name should come from the condition column (and be e.g. `24hpf`).

The two treps from it would both be called `24hpf_rep1 bw`, since the one is chosen from descriptive from the brep.

Issue #593, opened 2020-11-30
BUG: [scRNA: Trimming inconsistencies ]


tilschaef:
**Describe the bug**
Trimming of scRNA reads before running kallisto bustools does not produce expected results. Since we only trim the fastq that contains the reads, the number of total reads after trimming can vary between R1/R2.  

**To Reproduce**
The current implementation of the seq2science scRNA workflow can not reproduce the results from the [10xv3 pbmc notebook](https://colab.research.google.com/github/pachterlab/kallistobustools/blob/master/notebooks/kb_analysis_0_python.ipynb#scrollTo=bEN8ifXwEbki). This becomes especially apparent in the PCA plot which displays no cluster structure but appears random.

**Expected behavior**
The count tables produces by the scRNA workflow should yield comparable results to the 10xv3 pbmc notebook.  

**Possible solution**
Add a fastq matching rule after trimming to correct for differing read numbers between R1 and R2. This could be archived with (fastq-pair)[https://github.com/linsalrob/fastq-pair]

Maarten-vd-Sande:
Potential fix through pyfastx: https://github.com/lmdu/pyfastx/issues/22

However my current implementation is **slow**, so not attractive. Hopefully I am just being stupid.

PR #598, opened 2020-12-03
Sc rna fastq pair

tilschaef:
This commit fixes issue #593 by adding a fastq_pair rule after trimming of scRNA data. This ensures that R1/R2 fastqs contain the same number of reads before pseudo-alignment is performed. 

**What did change**
- Added fastq_pair rule and helper functions
- Added fastq_pair conda environment
- Updated scRNA schema
- Included trackhub schema in scRNA workflow
- Added kallistobus example settings

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+
 ## [0.3.1] - 2020-12-03
+
+### Fixed
 
 - Fixed scRNA trimming issue
 - Added fastq_pair
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
-## [0.3.1] - 2020-12-03
-
 ### Fixed
 
 - Fixed scRNA trimming issue
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
-
 ## [0.3.1] - 2020-12-03
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+## [0.3.1] - 2020-12-03
+
+- Fixed scRNA trimming issue
+- Added fastq_pair
+- Updated schemas
+- Added kallistobus examples
+
 ## [0.3.1] - 2020-11-16
 
 ### Added
--- a/seq2science/envs/qc.yaml
+++ b/seq2science/envs/qc.yaml
@@ -6,4 +6,3 @@ channels:
 dependencies:
   - bioconda::fastqc=0.11.8
   - bioconda::multiqc=1.9
-
--- a/seq2science/envs/qc.yaml
+++ b/seq2science/envs/qc.yaml
@@ -6,4 +6,4 @@ channels:
 dependencies:
   - bioconda::fastqc=0.11.8
   - bioconda::multiqc=1.9
-  - conda-forge:pretty_html_table=0.9.dev0
+
--- a/seq2science/envs/qc.yaml
+++ b/seq2science/envs/qc.yaml
@@ -6,3 +6,4 @@ channels:
 dependencies:
   - bioconda::fastqc=0.11.8
   - bioconda::multiqc=1.9
+  - conda-forge:pretty_html_table=0.9.dev0
--- a/seq2science/rules/quantification.smk
+++ b/seq2science/rules/quantification.smk
@@ -203,12 +203,20 @@ elif config["quantifier"] == "kallistobus":
             "../envs/fastq-pair.yaml"
         params:
             clean_dir=config.get("fastq_clean_dir"),
-            options=lambda wildcards, input: config.get("fastq_pair", "") if "-t" in config.get("fastq_pair", "") else config.get("fastq_pair", "") + " " +  "-t 100003"
+            options=config.get("fastq-pair",""),
+            tused=lambda wildcards, input: "true" if "-t" in config.get("fastq-pair", "") else "false"
         shell:
             """
             gunzip -c {input.r1} > {output.intermediates1[0]}
             gunzip -c {input.r2} > {output.intermediates1[1]}
-            fastq_pair {params.options} {output.reads}
+            if [ {params.tused} == true ]
+            then
+              opts="{params.options}"
+            else
+              opts="-p -t "$(wc -l {input.r1} | grep -Po '^\d+' | awk '{{print int($1/4)}}')
+            fi
+            fastq_pair $opts {output.intermediates1} 
+
             """
 
     rule kallistobus_count:
--- a/seq2science/rules/quantification.smk
+++ b/seq2science/rules/quantification.smk
@@ -166,33 +166,35 @@ elif config["quantifier"] == "kallistobus":
             {params.options} > {log} 2>&1
             """
 
-    def get_kallistobus_reads(wildcards):
-        reads = []
-        sample = wildcards.sample
+    def get_fastq_pair_reads(wildcards):
+        reads = dict()
         assert sampledict[sample]["layout"] == "PAIRED"
         read_id = get_bustools_rid(config.get("count"))
         #Determine mate for trimming
         if read_id == 0:
-            reads += expand("{fastq_clean_dir}/{{sample}}_clean_{fqext1}.{fqsuffix}.paired.fq", **config)
-            reads += expand("{fastq_clean_dir}/{{sample}}_clean_{fqext2}.{fqsuffix}.paired.fq", **config)
+            reads["r1"] = expand("{trimmed_dir}/{{sample}}_{fqext1}_trimmed.{fqsuffix}.gz", **config)
+            reads["r2"] = expand("{fastq_dir}/{{sample}}_{fqext2}.{fqsuffix}.gz", **config)
         elif read_id == 1:
-            reads += expand("{fastq_clean_dir}/{{sample}}_clean_{fqext1}.{fqsuffix}.paired.fq", **config)
-            reads += expand("{fastq_clean_dir}/{{sample}}_clean_{fqext2}.{fqsuffix}.paired.fq", **config)
-        else:    
+            reads["r1"] = expand("{fastq_dir}/{{sample}}_{fqext1}.{fqsuffix}.gz", **config)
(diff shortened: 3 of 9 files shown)


seq2science/rules/quantification.smk
```diff
@@ -166,22 +166,45 @@ elif config["quantifier"] == "kallistobus":
...
+            r2=expand("{fastq_clean_dir}/{{sample}}_clean_{fqext2}.{fqsuffix}.paired.fq", **config),
+            intermediates1=temp(expand("{fastq_clean_dir}/{{sample}}_clean_{fqext}.{fqsuffix}", **config)),             
+            intermediates2=temp(expand("{fastq_clean_dir}/{{sample}}_clean_{fqext}.{fqsuffix}{singles}.fq", **{**config,
+                                                                                                            **{"singles": [".single"]}}))                                                                               
+        priority: 1
+        conda:
+            "../envs/fastq-pair.yaml"
+        params:
+            clean_dir=config.get("fastq_clean_dir"),
+            options=config.get("fastq-pair")
```
siebrenf, line 201: {}.get("a") returns None. I think you need to set `default = ""`, like `options=config.get("fastq-pair", "")`?

seq2science/rules/quantification.smk
```diff
@@ -166,22 +166,45 @@ elif config["quantifier"] == "kallistobus":
...
+        shell:
+            """
+            gunzip -c {input.r1} > {params.clean_dir}/{wildcards.sample}_clean_R1.fastq 
+            gunzip -c {input.r2} > {params.clean_dir}/{wildcards.sample}_clean_R2.fastq
+            fastq_pair {params.options} {params.clean_dir}/{wildcards.sample}_clean_R1.fastq {params.clean_dir}/{wildcards.sample}_clean_R2.fastq
```
siebrenf, line 206: You hardcoded the R1 and R2 here.

siebrenf, line 206: If I got it correctly, you use `gunzip`  to make the intermediate files? If so, you could replace it with 
```
gunzip -c {input.r1} > {output.intermediates1}
gunzip -c {input.r2} > {output.intermediates2}
fastq_pair {params.options} {output.intermediates1} {output.intermediates2}
```
saves you a param as well!

seq2science/rules/quantification.smk
```diff
@@ -166,22 +166,45 @@ elif config["quantifier"] == "kallistobus":
...
+            reads["r1"] = expand("{fastq_dir}/{{sample}}_{fqext1}.{fqsuffix}.gz", **config)
+            reads["r2"] = expand("{trimmed_dir}/{{sample}}_{fqext2}_trimmed.{fqsuffix}.gz", **config)
+        else:
             raise NotImplementedError
-
         return reads
+        
+    rule fastq_pair:    
+        """
+        Example
```
siebrenf, line 186: nitpick: i have no idea what the rule is doing. is it trimming, is it preparing the other half of the sample for trimming?  What is `get_bustools_rid`?

seq2science/rules/quantification.smk
```diff
@@ -166,22 +166,45 @@ elif config["quantifier"] == "kallistobus":
             {params.options} > {log} 2>&1
             """
 
-    def get_kallistobus_reads(wildcards):
-        reads = []
-        sample = wildcards.sample
+    def get_fastq_pair_reads(wildcards):
+        reads = dict()
         assert sampledict[sample]["layout"] == "PAIRED"
```
siebrenf, line 171: nitpick (its not even in this PR, sorry): add an error message if the assert is tripped. Something like
`assert sampledict[sample]["layout"] == "PAIRED", "we only support paired-end scRNA-seq"`

seq2science/rules/quantification.smk
```diff
@@ -166,22 +166,45 @@ elif config["quantifier"] == "kallistobus":
...
+        
+    rule fastq_pair:    
+        """
+        Example
+        """
+        input:
+            unpack(get_fastq_pair_reads)
+        output:
+            r1=expand("{fastq_clean_dir}/{{sample}}_clean_{fqext1}.{fqsuffix}.paired.fq", **config),
+            r2=expand("{fastq_clean_dir}/{{sample}}_clean_{fqext2}.{fqsuffix}.paired.fq", **config),
```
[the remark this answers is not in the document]
siebrenf: P.S. we already make the assumption that fqext1 and fqext2 are ALWAYS lexicographically sorted, so they should always be found as R1 > R2.

siebrenf (changes requested):
Mostly of little details here, but there are 2 important remarks (unless I misread, which happens a lot):
- config.get() needs a default
- you hardcoded the fqext

See the comment for more details :)

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+## [0.3.1] - 2020-12-03
+
+- Fixed scRNA trimming issue
+- Added fastq_pair
+- Updated schemas
+- Added kallistobus examples
```
Maarten-vd-Sande, line 16: Should be under unreleased!

seq2science/envs/qc.yaml
```diff
@@ -6,3 +6,4 @@ channels:
 dependencies:
   - bioconda::fastqc=0.11.8
   - bioconda::multiqc=1.9
+
```
Maarten-vd-Sande, line 9: :eyes:

seq2science/rules/quantification.smk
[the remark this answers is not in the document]
Maarten-vd-Sande: R1 and R2 are used by the rule downstream, so I think this is nice and verbose this way

    input:
         r1=rules.fastq_pair.output.r1,
         r2=rules.fastq_pair.output.r2

seq2science/rules/quantification.smk
Maarten-vd-Sande: Maybe the -t can be derived dynamically, something like this?

```
wc -l {input.r1} | grep -Po "^\d+" | awk '{print int($1/4)}'
```

If you want to be really fancy you could make it so that if the user specifies a `-t` you use theirs, if they don't you use the dynamicly derived one.

Untested:

```
options=lambda wildcards, input: config.get("fastq-pair", "") if "-t" in config.get("fastq-pair", "") else config.get("fastq-pair", "") + " " + f"-t \"$(wc -l {input.r1} | grep -Po "^\d+" | awk '{{print int($1/4)}}'\""
```

Needs some cleaning up :yum:

seq2science/workflows/scrna_seq/config.yaml
```diff
@@ -18,7 +16,13 @@ technical_replicates: merge
...
-    #count: '-x 10XV2 --verbose --lamanno'
+    ## Quantification example ##
+    # count: '-x 0,8,16:0,0,8:1,0,0 --h5ad --verbose'
+    # count: '-x 10XV3 --h5ad --verbose'
+
+    ## Velocity example ##
+    # ref: '--workflow lamanno'
+    # count: '-x 0,8,16:0,0,8:1,0,0 --h5ad --verbose --workflow lamanno'
+    # count: '-x 10XV3 --verbose --h5ad --workflow lamanno'
+
```
Maarten-vd-Sande, line 27: What happens if you leave both uncommented?

Maarten-vd-Sande (changes requested):

CHANGELOG.md
tilschaef: Added fixed section under unreleased

seq2science/rules/quantification.smk
tilschaef: Added description to fastq_pair rule.  You can find `get_bustools_rid` in `utils.py`. Basically, it tells Fastp which FASTQ should be trimmed and which one contains the scRNA barcoded. You can specify that via Kallisto bustool settings.

seq2science/rules/quantification.smk
tilschaef: Added error message

seq2science/rules/quantification.smk
tilschaef: It should be a bit more dynamic since I set the filenames based on the output.

seq2science/rules/quantification.smk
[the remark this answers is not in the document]
tilschaef: @siebrenf It can be reduced if I use `fqext` indeed. Since its lexicographically sorted. If I reference it via `rules.fastq_pair_output_reads` in `kallistobus_count` ,It should not cause any troubles I would assume.

seq2science/rules/quantification.smk
```diff
@@ -166,22 +166,50 @@ elif config["quantifier"] == "kallistobus":
...
+        output:
+            reads=expand("{fastq_clean_dir}/{{sample}}_clean_{fqext}.{fqsuffix}.paired.fq", **config),
+            intermediates1=temp(expand("{fastq_clean_dir}/{{sample}}_clean_{fqext}.{fqsuffix}", **config)),             
+            intermediates2=temp(expand("{fastq_clean_dir}/{{sample}}_clean_{fqext}.{fqsuffix}{singles}.fq", **{**config,
+                                                                                                            **{"singles": [".single"]}}))                                                                               
+        priority: 1
+        conda:
+            "../envs/fastq-pair.yaml"
+        params:
+            clean_dir=config.get("fastq_clean_dir"),
```
siebrenf, line 205: this param is now obsolete

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
...
+## [0.3.1] - 2020-12-03
+
+### Fixed
+
+- Fixed scRNA trimming issue
+- Added fastq_pair
+- Updated schemas
+- Added kallistobus examples
+
 ## [0.3.1] - 2020-11-16
```
siebrenf, line 21: still something weird happening here. I think you can just remove line 12?

siebrenf (approved):
thanks for the extra details :)

seq2science/workflows/scrna_seq/config.yaml
tilschaef: Then it will probably fail. I don't think its possible to set sane defaults for Kallisto bus since it highly depends on the technology you use and workflow (quantification, velocity). For now, it performs 10xv3 scRNA quantification by default.

seq2science/rules/quantification.smk
tilschaef: Good one. Now it determines -t dynamically or derives them from the config settings (if present)

CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+## [0.3.1] - 2020-12-03
```
Maarten-vd-Sande, line 11: this line should still be removed. It should be udner unreleased

Maarten-vd-Sande (changes requested):

Maarten-vd-Sande (approved):

PR #599, opened 2020-12-04
Idr fix

Maarten-vd-Sande:
IDR narrowpeaks did not work with combine_peaks. Now does (by hacking the summit in the middle of the peak). I had some issues with the macs2 and khmer, and adding setuptools fixed it. Not sure why or what was happening there..

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/envs/macs2.yaml
+++ b/seq2science/envs/macs2.yaml
@@ -7,3 +7,4 @@ dependencies:
   - bioconda::macs2=2.2.7
   - bioconda::khmer=2.0
   - conda-forge::jq=1.6
+  - anaconda::setuptools=v50.3.1
--- a/seq2science/rules/peak_calling.smk
+++ b/seq2science/rules/peak_calling.smk
@@ -283,6 +283,9 @@ if "condition" in samples:
             """
             Combine replicates based on the irreproducible discovery rate (IDR). Can only handle two replicates.
             For more than two replicates use fisher's method.
+
+            When combining narrowpeak files we make the q-score zero, and the peak summit in the middle. This is
+            necessary for later downstream processing, but are not the "true" values.
             """
             input:
                 get_idr_replicates,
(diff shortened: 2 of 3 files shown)


seq2science/envs/macs2.yaml
```diff
@@ -7,3 +7,4 @@ dependencies:
   - bioconda::macs2=2.2.7
   - bioconda::khmer=2.0
   - conda-forge::jq=1.6
+  - anaconda::setuptools=v50.3.1
```
siebrenf, line 10: wait wut?

seq2science/rules/peak_calling.smk
```diff
@@ -283,11 +283,15 @@ if "condition" in samples:
             """
             Combine replicates based on the irreproducible discovery rate (IDR). Can only handle two replicates.
             For more than two replicates use fisher's method.
+
+            When combining narrowpeak files we make the q-score zero, and the peak summit in the middle. This is
```
siebrenf, line 287: typo and a bit vague. Did you mean "...take the q-score, and the center of the peak summit."

siebrenf (approved):

seq2science/envs/macs2.yaml
Maarten-vd-Sande: Yeah I don't know. I got some problems on my own computer with the environment. I googled and the problem was supposed to be caused by an old setuptools. I don't see what caused the problem, but my idea was that this doesn't hurt.

seq2science/rules/peak_calling.smk
Maarten-vd-Sande: Nope. IDR makes the qscore and summit both -1, that destroys some downstream tools. So I decided to put the summit in the middle, and the q-score as zero. I'll update the text to be more clear

PR #600, opened 2020-12-07
biological_replicates and technical_replicates

Maarten-vd-Sande:
Since I was the one originally opposing the long names, it only seems fair I do this one :yum: 

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
@siebrenf this is ready. Maybe we should keep this here a bit longer? We can release it maybe together with rerunning and scRNA? Or too much big changes at the same time, so if bugs appear we won't know what is causing it?

seq2science/rules/peak_counts.smk
```diff
@@ -140,7 +140,7 @@ def get_coverage_table_replicates(file_ext):
 
 rule coverage_table:
     """
-    Use gimmemotif's coverage_table to generate a cpunt table with the nr of reads
+    Use gimmemotif's coverage_table to generate a count table with the nr of reads
```
siebrenf, line 143: 🎉

siebrenf (changes requested):
scripts still need some attention. sec.

siebrenf (approved):
I think that's it.

Yea, we should run a few more test runs with this. I'll rerun Angela's (RNA-seq & ATAC-seq) data to compare it! (she should have it ~ next week)

Maarten-vd-Sande:
thanks for the catches, I really though I had them all (those names are everywhere :yum: )

PR #608, opened 2020-12-10
star_align log

siebrenf:
Simple name change as the directory looked like a file.

More cleanup possible: The STAR output directory only contains the SJ.out.tab.

The SJ.out.tab is used in the 2-pass system for _de novo_ alignment, which we don't support, but might be interesting to others.
We could leave it as-is, or lose it, or leave it in the log directory*. With the latter two options we can make the output.dir a temporary directory. What do you think?

*STAR itself considers the file as a Log file


Maarten-vd-Sande:
I think it is fine as is, but I don't really know what all the options are with *de novo* stuff. When using the `alignment` workflow it might be interesting to have this as output?

Maarten-vd-Sande (approved):

siebrenf:
it could be interesting for differential transcript analysis, so its fine to keep it yea!

PR #610, opened 2020-12-10
Update get_genome_annotation.py

siebrenf:
A description of the problem that is being solved, or which new feature is implemented. Please link to other relevant issues/pr if relevant

**What did change**
Explain clearly and concisely what you changed. Why did you take certain design decisions?

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/scripts/get_genome.py
+++ b/seq2science/scripts/get_genome.py
@@ -4,6 +4,7 @@ Script to download genome
 import contextlib
 
 import genomepy
+from snakemake.logging import logger
 
 
 with open(snakemake.log[0], "w") as log:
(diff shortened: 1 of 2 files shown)


Maarten-vd-Sande:
I am not sure if this is the cause of the problem. I think it happens because genomepy has trouble connecting to some server(s)

https://github.com/vanheeringen-lab/genomepy/blob/master/genomepy/provider.py#L72

siebrenf:
I might have mixed up the two errors we've been having with this section. 

That said, this error message will only display if the rules run, and while running, are requested to download from an unsupported provider. So it might still help with that other issue...

Maarten-vd-Sande:
I usually avoid `seq2science clean` since it also removes all the installed environments, and that takes quite some time. :upside_down_face:

siebrenf:
fair. we could instead delete the genomepy pickle in this if-statement, and provide explanation. I'll make something better when I have the time :)

Maarten-vd-Sande:
No, that will introduce lots of edge-case bugs (e.g. downloading multiple genomes). Please don't! :yum: 

I don't think I ever had problems with a corrupted genomepy+seq2science cache.

Maarten-vd-Sande:
@siebrenf maybe something like this? This also should not print the exception anymore, it should just add it to the log (as it is supposed to do)

Maarten-vd-Sande (approved):
very good, the smartest solution, only the best :+1:

PR #615, opened 2020-12-14
explain more

Maarten-vd-Sande:
I got questions about how peaks were merged and count table was made, so here is a bit more explanation. I decided to also add testing functionality for these settings.

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/tests/dag_tests.sh
+++ b/tests/dag_tests.sh
@@ -464,6 +464,31 @@ if [ $1 = "rna-seq" ]; then
   test_ran=1
 fi
 
+if [ $1 = "explain" ]; then
+  yes | seq2science init download-fastq
+  seq2science explain download-fastq
+
+  yes | seq2science init alignment
+  seq2science explain alignment
+
+  yes | seq2science init atac-seq
+  seq2science explain atac-seq
+
+  yes | seq2science init chip-seq
+  seq2science explain chip-seq
+
+  yes | seq2science init rna-seq
+  seq2science explain rna-seq
+
+  yes | seq2science init scrna-seq
+  seq2science explain scrna-seq
+
+  yes | seq2science init scatac-seq
+  seq2science explain scatac-seq
+
+  test_ran=1
+fi
+
 # check if any test has run
 if [ -z "$test_ran" ]; then
   printf "\nunrecognized input: ${1}\n"; exit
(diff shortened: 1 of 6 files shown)


tests/dag_tests.sh
```diff
@@ -464,6 +464,31 @@ if [ $1 = "rna-seq" ]; then
...
+
+  yes | seq2science init scrna-seq
+  seq2science explain scrna-seq
+
+  yes | seq2science init scatac-seq
+  seq2science explain scatac-seq
+
+  test_ran=1
+fi
+
```
siebrenf, line 491: This should probably catch stdout and make sure it contains at least one expected line! If the function fails due to an error in the messages, it outputs a truncated message.

siebrenf: something like this perhaps?

```
function assert_stdout {
  # store any lines containing the argument in val
  val=$(cat stdout.log | grep $1);

  # if $val is empty, exit
  if [ -z "$val" ]; then
    printf "\nExplain failed. Exiting.\n\n" && exit 1;
  fi
  :
}

seq2science explain download-fastq > stdout.log
assert_stdout "Public samples"
```

siebrenf (commented):
I'm all in for more fast tests, and more explaining. 

One question tho (see the comment)

tests/dag_tests.sh
Maarten-vd-Sande: Oh really? I didn't check it, I assumed it would fail because of the exit status..

siebrenf: This happens if there is an error in the explain messages, like an incorrect reference. IIRC it doesn't trigger an error.

Maarten-vd-Sande: https://github.com/vanheeringen-lab/seq2science/blob/217bcc80cf75918c904914a41dd3238b260b38b5/seq2science/rules/configuration_explain.smk#L18

I think we should not make this a get. I think we should crash when the message isn't present :thinking: . Let me try some stuff. Unless that means *all* rules need a message, but I doubt it. Since we call this func.

Maarten-vd-Sande:
@siebrenf now it should work :)

siebrenf (approved):

PR #627, opened 2021-01-17
min and max insert size

Maarten-vd-Sande:
Some atac-seq guides recommend only using fragments from the nucleosome free region (insert size smaller than ~150-200 bp). This is now implemented. The implementation is a bit mehh, since we do want to get the insert size plot, so we split the sieve result into two.

1. After all the sieving is done, except the insert size sieving. This is used for the multiqc plot insert size plot
2. After all the sieving, including insert size sieving. This is used for the remainder of the workflow

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
proof of concept

image

image

seq2science/rules/bam_cleaning.smk
```diff
@@ -241,6 +262,21 @@ rule samtools_index:
         samtools index {params} {input} {output}
         """
 
+rule sam2bam:
```
siebrenf, line 265: nothing wrong with this rule, but any error with bam files in the future will point to this rule!

seq2science/rules/configuration_workflows.smk
```diff
@@ -62,6 +62,16 @@ if config.get("peak_caller", False):
         else:
             config["macs2_types"].extend(["summits.bed", "peaks.narrowPeak"])
 
+# make sure that both maximum and minimum insert sizes are existing when one of them is used
+if config.get("min_template_length") and not config.get("max_template_length"):
+    config["max_template_length"] = 1_000_000_000
```
siebrenf, line 67: woah! Did not know you could write long numbers like that!

tests/run_tests.sh
```diff
@@ -163,10 +163,10 @@ if [ $1 = "atac-seq" ]; then
 #  seq2science run atac-seq --cores $CORES --configfile tests/alignment/remote_genome_n_sample.yaml --snakemakeOptions config={aligner:bowtie2}
 
   printf "\natac-seq - multiqc report\n"
-  seq2science run atac-seq --cores $CORES --configfile tests/alignment/remote_genome_n_sample.yaml --snakemakeOptions config={aligner:bowtie2,create_qc_report:True,custom_genome_extension:tests/tinydata/tinyERCC92.fa} show_failed_logs=True
+  seq2science run atac-seq --cores $CORES --configfile tests/alignment/remote_genome_n_sample.yaml --snakemakeOptions config={aligner:bowtie2,create_qc_report:True,custom_genome_extension:tests/tinydata/tinyERCC92.fa,max_template_length:150} show_failed_logs=True
```
siebrenf, line 166: change one of these into `min_template_length:30`, just to have a test with it too.

siebrenf (approved):

tests/run_tests.sh
siebrenf: i figured you meant to do that already and/or dont mind me, so its done

seq2science/rules/bam_cleaning.smk
Maarten-vd-Sande: Yes, maybe I will put an if statement around it

tests/run_tests.sh
Maarten-vd-Sande: thanks :)

PR #629, opened 2021-01-18
update RNA-seq QC

siebrenf:
**What's new?**
add all the nice QC features into RNA-seq where applicable

**Checklist**
- [x] I updated the CHANGELOG


seq2science/rules/qc.smk
```diff
@@ -693,7 +693,8 @@ def get_alignment_qc(sample):
     # add samtools stats
     output.append(f"{{qc_dir}}/markdup/{{{{assembly}}}}-{sample}.samtools-coordinate.metrics.txt")
     output.append(f"{{qc_dir}}/samtools_stats/{{aligner}}/{{{{assembly}}}}-{sample}.samtools-coordinate.samtools_stats.txt")
-    if sieve_bam(config):
+    if sieve_bam(config) and \
+            not (get_workflow() == "rna_seq" and config.get('quantifier') == 'salmon'):
```
Maarten-vd-Sande, line 697: Is this correct?

siebrenf: Yep! It will add post-sieve data, unless Salmon is used for quantification. 

Why: if Salmon is used, the sieving does not effect the expression values, so adding it to the MultiQC is confusing (other quantifiers do use the sieved bams).

Maarten-vd-Sande: Ah ik I understand. Maybe add a comment about it? 😇

Maarten-vd-Sande (approved):

PR #636, opened 2021-02-05
resource parsing profile

Maarten-vd-Sande:
Small bug fix in profile parsing where the resources were loaded as a list of strings instead of a dict.

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
ah, so that's what was causing the profile cores to be missed!

Issue #606, opened 2020-12-09
FR: separate configs for quantification and alignment


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
When not using an aligner (only a quantifier), e.g. salmon or kallistobus, the aligner and bam sorting are still printed as the config. This is confusing, and should not happen.

PR #639, opened 2021-02-05
hide aligner config stuff

siebrenf:
see #606 

Aligners are required for quantifiers htseq and featurecounts, but also if a trackhub is being made.
Because of this, its easier to just remove the keys from the config. 

@tilschaef, will this simple check suffice for scRNA too?

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+### Changed
+
+- slimmed down the config printed at the start of a run
+
 ### Added
 
 - Options to filter bams by minimum and maximum insert sizes (added to config of bulk/sc atac)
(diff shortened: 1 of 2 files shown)


tilschaef:
@siebrenf, look alright to me. Just to clarify, it checks if the quantifier is set to Kallisto/Salmon and if not, would remove all the alignment related keys?

Maarten-vd-Sande (approved):

tilschaef (approved):

siebrenf:
Almost. If quantifier is Kallisto/Salmon and no trackhubs is made, then remove the aligner stuff. 
For RNA-seq, this means aligners are unused. I guess the same goes for scRNA-seq, but that's why I asked for your review :)

Maarten-vd-Sande:
Oh ehm my bad, premature merge?

siebrenf:
the PR only removes keys from the logger. not internally. No harm done even if it was too soon :)

Issue #619, opened 2020-12-18
FR: better estimation memory usage of rules


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**
Slurm on cn45 kills a process if it uses more memory than requested, that's bad!

**Describe the solution you'd like**
Better memory description per rule. We should map which rules are problematic first.

Maarten-vd-Sande:
With the newest snakemake-profiles we can add time as a resource, which helps the scheduler :)

```
resources:
    time="1-12:00:00
```

The more we specify, especially short rules, the sooner they get scheduled when it is busy on the scheduler :tada:

siebrenf:
woops, wrong issue

siebrenf:
Does snakemake allocate a default amount of memory to a rule if resources are unset?
I've found we can set a default, but I'm not sure if there is one if you didn't...

Maarten-vd-Sande:
I don't think we set a default, you mean like this?
https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#standard-resources

Yeah makes sense. I am setting it now with the profile, but it is better to do it like that.

siebrenf:
We don't set a default AFAIK. That link indeed shows how we could do this, so it might help solve this issue.

My question however, was slightly different:

What exactly causes Slurm to complain? is it that:
- we set mem_gb, and the rule used more
- we did not set mem_gb, snakemake then claims a default amount, and the rule used more
- snakemake takes up too much RAM at the same time at any point (meaning we should add ~usage to each rule, and add an upper limit via the CLI?)

I'm adding some expected values now, but I'm not sure if it might do more harm! 😨

PR #640, opened 2021-02-05
mem per rule

siebrenf:
see #619 

`rule computematrix` apparently takes >47 GB according to my one benchmark.


Maarten-vd-Sande:
My guess is that the memory used by _computematrix_ depends a lot on the number of input files (and assembly size). I could scale it. I will check if I have some logs of it

Maarten-vd-Sande (approved):
I will do a benchmark soon for human, with 1, 2, 4, 8, and 16 samples. Just to see how computematrix scales

Maarten-vd-Sande:
Merging this because I am too lazy for benchmarks (at the moment)

PR #641, opened 2021-02-05
Differential peaks

siebrenf:
Add deseq2 to ATAC-seq and ChIP-seq

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
Annoying now that `biological_replicates` has a underscore now.. Tried for 10 minutes to fix the deseq2 script, but idk how R works

siebrenf:
right, I forgot about this when my internet broke! I'll fix this after the ANANSE stuff is done!

siebrenf:
i think this allows us to have columns with underscores (as long as there is more than just underscores :p)

but we now no longer support `column_group` contrasts (which is minor, since `column_group_all` is still accepted)

siebrenf:
DAG is working. Code wise it looks OK. Can't connect to our server atm, so here are the TODOs:

- test some RNA-seq data /w batch effect, quantified with salmon, with and without technical replicates
- test some ATAC/ChIP-seq data, with and without technical replicates
- save (part of) the counts tables as example data for tests
- add/update tests

docs/content/workflows/rna_seq.md
```diff
@@ -128,22 +128,78 @@ To do so, you can add the color by name (google "matplotlib colors" for the opti
...
+|---|---|---|
+|sample_1|hg38|control|
+|sample_2|hg38|control|
+|sample_3|hg38|treatmentA|
+|sample_4|hg38|treatmentA|
+|sample_5|hg38|treatmentB|
+|sample_6|hg38|treatmentB|
+
+</td></tr>
+</table>
```
Maarten-vd-Sande, line 178: Not sure if this works. Did you test if this build properly?

seq2science/rules/configuration_generic.smk
```diff
@@ -114,23 +114,24 @@ if len(errors):
...
-# (sample names if replicates are not found/applicable)
-# 2) drop column if it is identical to the replicate/sample column, or if not needed
+#    (sample names if replicates are not found/applicable)
+# 2) drop column if it provides no information
+#    (renamed in case it's used in a DE contrast)
 if 'technical_replicate' in samples:
     samples['technical_replicate'] = samples['technical_replicate'].mask(pd.isnull, samples['sample'])
     if len(samples['technical_replicate'].unique()) == len(samples['sample'].unique()) or config.get('technical_replicates') == 'keep':
-        samples = samples.drop(columns=['technical_replicate'])
+        samples.rename(columns={'technical_replicate': '_trep'}, inplace=True)
```
Maarten-vd-Sande, line 125: why are we renaming to these names?

docs/content/workflows/rna_seq.md
siebrenf: [It works in git md](https://github.com/vanheeringen-lab/seq2science/blob/a8964977cb5d9d18b6f3a22cb4eebe1fff8567c3/docs/content/workflows/rna_seq.md#conditions), any chance it then doesn't work in the github action?

Maarten-vd-Sande: Yes, because of markdown also having many flavours, and we convert it to html with some obscure sphinx integration :shrug:

seq2science/rules/configuration_generic.smk
siebrenf: testing showed me an edge case:
- your DE contrast is `biological_replicate_group2_group1`
- your config option `biological_replicates` is set to `keep`
- s2s would then drop this column internally (less keyword columns = simpler logic, so I'd keep it this way)

But now your DE contrast column is gone. So the DE contrast check fails.
To fix it I rename the column instead of dropping it, and we check these for DE contrasts too ([here](https://github.com/vanheeringen-lab/seq2science/blob/a8964977cb5d9d18b6f3a22cb4eebe1fff8567c3/seq2science/util.py#L248)).

In DESeq2.R, we reload the samples.tsv file, so the `biological_replicate` column is still there. Because the check was already performed in s2s, we know the contrast is valid.

docs/content/workflows/rna_seq.md
siebrenf: That did indeed not work in sphinx. Good catch!

siebrenf:
That should be it! 

Do we need to add `--skip-rerun` to (all) tests? Can do that in another PR in any case.

Finally, its in the changelog already, but I moved `onehot_peaks.tsv` to the `counts_dir` for consistency. Are you OK with that too?

Maarten-vd-Sande (approved):
Looks good, awesome new feature :tada: :muscle: 

I would add it to the chip/atac seq docs though, otherwise no one would find out. Also add to the default config a deseq2 entry?

PR #642, opened 2021-02-17
tab completion seq2science

Maarten-vd-Sande:
This is the groundwork to allow tab-completion for seq2science. For it to work you need to run / add the following line to your .bashrc: `eval "$(register-python-argcomplete seq2science)"`. However it is not nice to add it to ~/.bashrc ourselves, or to ask others to do it. Even more so because we only want it to work in the relevant environment! The oracle (dpryan) advised to use activation scripts: https://docs.conda.io/projects/conda-build/en/latest/resources/activate-scripts.html

This means that we need to update our recipe for this to work _out of the box_, and this might prove difficult, but... It is very cool :sunglasses: 

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 ### Added
 
 - Options to filter bams by minimum and maximum insert sizes (added to config of bulk/sc atac)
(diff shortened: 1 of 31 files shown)


Maarten-vd-Sande:
This works! :muscle: 

Updated recipe here: https://github.com/bioconda/bioconda-recipes/pull/26826

siebrenf (approved):

PR #643, opened 2021-02-19
slighly better samples.tsv filtering

siebrenf:
if the number of unique names in the replicate columns are the same to the samples column, there is no merging happening. So we dont need the columns.


Maarten-vd-Sande:
You mean in the case when all there is 1 trep per technical replicate right? I guess that's indeed slightly better :smile: 

I think it is `technical_replicateS`

Maarten-vd-Sande (approved):

siebrenf:
> I think it is `technical_replicateS`

that's the config option. I referenced the wrong column (fixed now). Your remark did led me to #647!

Maarten-vd-Sande:
@siebrenf this is stupid :see_no_evil: , should we make everything with or without S?

PR #646, opened 2021-02-22
Update requirements.yaml

siebrenf:
should improve genome install speed & stability a bit

### DIFF
--- a/requirements.yaml
+++ b/requirements.yaml
@@ -8,7 +8,7 @@ dependencies:
   - bioconda::snakemake=5.32.0
   - pkgs/main::conda=4.9.2
   - conda-forge::mamba=0.7.10
-  - bioconda::genomepy=0.9.1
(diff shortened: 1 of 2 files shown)


Maarten-vd-Sande (approved):
just be careful about the dependency stuff that pops up from it :)

siebrenf:
genomepy 0.9.3 requires tqdm >= 4.49, 
pysradb (0.11.1) == 4.51. 

So now I'm updating some packages. The new pysradb should work nicely with genomepy, and the other changes are hotfix version upgrades anyway...

siebrenf:
pysradb now returns `pd.NA` instead of "N/A". Apparently, working with `pd.NA` is a little tricky, cause pandas made a note of it in their docs. I found that out halfway through these commits :p

Maarten-vd-Sande (approved):
Seems fine :smile:

PR #652, opened 2021-02-26
dupRadar

siebrenf:
**What's new?**
add dupRadar to the RNA-seq MultiQC

**Checklist**
- [ ] I updated the CHANGELOG


siebrenf:
TODO:
- extract sample strandedness (they complain if you say paired ended, but the sample is single ended, so why do I have to tell them!?)
- benchmark RAM usage
- stop removing duplicate reads in RNA-seq
- add test
- update docs & changelog

siebrenf:
dups are removed here, but its the thought that counts!

GRCz11-GSM1483769_mqc

siebrenf:
afbeelding
(seems our test data is quite bad!)

- the rule takes <500 MB ram with these samples. Too low to care for resource logging I guess.
- dupRadar will run in the RNA-seq MultiQC tests, so it's covered.
- since the module isn't in MultiQC, each png is added as a separate file.
- since each file has it's sample name, I cant actually add any config options. I've added the "dupRadar" prefix to group them together:
afbeelding

seq2science/rules/configuration_explain.smk
```diff
@@ -73,7 +73,7 @@ else:
                       end=" and finally were tn5 bias shifted by seq2science." if config.get("tn5_shift", 0) > 0 else "."),
         # "samtools_sort": "Bam files were sorted with samtools v@samtools[samtools].",
         "sambamba_sort": "Bam files were sorted with sambamba v@sambamba[sambamba] (https://doi.org/10.1093/bioinformatics/btv098).",
-        "mark_duplicates": "Afterwards, duplicate reads were removed with picard MarkDuplicates v@picard[picard] (http://broadinstitute.github.io/picard).",
+        "mark_duplicates": "Afterwards, duplicate reads were " + ("removed" if "REMOVE_DUPLICATES=true" in config.get("markduplicates", "") else "marked") + " with picard MarkDuplicates v@picard[picard] (http://broadinstitute.github.io/picard).",
```
Maarten-vd-Sande, line 76: I am not sure, but I think most/some tools will not consider reads marked as duplicates. But from the comments with pictures seems that is not an issue actually :thinking:

Maarten-vd-Sande (approved):
My general idea of the multiqc report is that it starts at the beginning of the analysis (e.g. trimming), and ends with the later steps (perhaps other way around makes more sense though?). Anyways, ideally these plots would be added at the bottom. 

The multiqc config is basically 1 big f string, so you should be able to add them. Take a look at e.g. `macs2_img1`.

seq2science/rules/configuration_explain.smk
siebrenf: dupRadar actually needs reads marked

Maarten-vd-Sande: yeah makes sense actually :see_no_evil:

PR #655, opened 2021-03-01
Snakemake newer version

Maarten-vd-Sande:
Let's see. It's good to try to keep up to date. 

One of the problems we have with newer snakemake versions is that it ignores some RNA-seq steps.

**Checklist**
- [ ] I updated the CHANGELOG


Maarten-vd-Sande:
Changing the `run:` directive to `script:` directive does not solve this :sob:

siebrenf:
the checkpoint is used in 5 tests, and fails consistently in 4!
- fails in 3 gene counts tests
- works in trackhub
- fails multiQC

siebrenf (approved):

PR #661, opened 2021-03-09
fix

### LINKED ISSUE #660: FR: fastq naming convention
**Is your feature request related to a problem? Please describe.**
Fastq files are currently made with a "new" extension: https://support.illumina.com/help/BaseSpace_OLH_009008/Content/Source/Informatics/BS/NamingConvention_FASTQ-files-swBS.htm ending with 001

Should we change our fqext settings? Or should we auto-detect this somehow? :thinking:

siebrenf:
fix #660 by gzipping and renaming local fastqs.
We might not need to gzip, and we could symlink instead. What do you think Maarten?

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
Why can't we do something like `renamefastq_PE`? Best to have it as rules I guess.

gzipping is not necessary at all, since trimmers work with and without zipped fastqs. Also that will take a while, so s2s is just hanging waaay to long before it actually starts

Maarten-vd-Sande:
@siebrenf Sorry I kinda forgot about this. What's the status on this, some tests are failing? I guess only accepting gzipped is fine, we can't be expected to cover **everything**

siebrenf:
its... almost working :p
image

I'll fix this soon<sup>tm</sup>

tests/dag_tests.sh
```diff
@@ -63,8 +63,8 @@ if [ $1 = "alignment" ]; then
   printf "\ndownload default\n"
   seq2science run download-fastq -n --configfile tests/$WF/default_config.yaml --snakemakeOptions quiet=True | tee tests/local_test_results/${1}_dag
   assert_rulecount $1 runs2sample 3
-  assert_rulecount $1 ena2fastq_SE 5
-  assert_rulecount $1 ena2fastq_PE 1
+  assert_rulecount $1 'ena2fastq_SE|sra2fastq_SE' 5
+  assert_rulecount $1 'ena2fastq_PE|sra2fastq_PE' 1
```
Maarten-vd-Sande, line 67: Nice, but also not nice. ENA was down couple days, so that is annoying. However if something gets broken in ena downloading we won't realize

seq2science/rules/qc.smk
```diff
@@ -670,23 +670,26 @@ rule multiqc:
...
             # therefore we can not just always start trimming paired-end even though
             # the samples are paired-end (ish)
             read_id = get_bustools_rid(config.get("count"))
             if read_id == 0:
-                return expand([f"{{qc_dir}}/fastqc/{sample}_R1_fastqc.zip",
-                               f"{{qc_dir}}/fastqc/{sample}_R1_trimmed_fastqc.zip",
-                               f"{{qc_dir}}/trimming/{sample}_R1.{{fqsuffix}}.gz_trimming_report.txt"],
+                return expand([f"{{qc_dir}}/fastqc/{sample}_{{fqext1}}_fastqc.zip",
+                               f"{{qc_dir}}/fastqc/{sample}_{{fqext1}}_trimmed_fastqc.zip",
+                               f"{{qc_dir}}/trimming/{sample}_{{fqext1}}.{{fqsuffix}}.gz_trimming_report.txt"],
```
Maarten-vd-Sande, line 687: :see_no_evil:

tests/dag_tests.sh
siebrenf: Samples will always be downloaded from ENA if it is up. We'd need some samples available only from SRA if you want to properly test both sets of rules. Ill make a separate issue for this.

PR #662, opened 2021-03-10
only log real runs

siebrenf:
Dry runs don;t need to be logged.

### DIFF
--- a/seq2science/cli.py
+++ b/seq2science/cli.py
@@ -310,7 +310,8 @@ def _run(args, base_dir, workflows_dir, config_path):
 
     # run snakemake/seq2science
     #   1. pretty welcome message
-    setup_seq2science_logger()
+    if not parsed_args["dryrun"]:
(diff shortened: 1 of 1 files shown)


Maarten-vd-Sande (approved):

seq2science/cli.py
```diff
@@ -546,13 +545,14 @@ def resource_parser(parsed_args):
         parsed_args["resources"]["mem_gb"] = round(mem)
 
 
-def setup_seq2science_logger():
-    seq2science_logfile = os.path.abspath(
-        "seq2science."
-        + datetime.datetime.now().isoformat().replace(":", "")
-        + ".log"
-    )
+def setup_seq2science_logger(dryrun):
```
Maarten-vd-Sande, line 548: Perhaps pass parsed_args instead of dryrun? Doesn't really matter though

Maarten-vd-Sande (approved):

PR #663, opened 2021-03-12
multiqc 1.10

Maarten-vd-Sande:
I added macs2 number of peaks in a narrowpeak file to multiqc, and this has now been released :tada:. Unfortunately the scRNA QC PR was not yet merged (https://github.com/ewels/MultiQC/pull/1341), so that will happen later.

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/envs/fastqc.yaml
+++ b/seq2science/envs/fastqc.yaml
@@ -1,8 +1,7 @@
-name: qc
+name: fastqc
 channels:
   - conda-forge
   - bioconda
   - defaults
 dependencies:
   - bioconda::fastqc=0.11.8
-  - bioconda::multiqc=1.9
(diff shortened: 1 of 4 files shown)


siebrenf (approved):
why split fastqc and multiqc? Considering their contents it may cause more hassle (fastqc needs to be updated in 2 places)

Maarten-vd-Sande:
Since we have also have `fastp` as a (default) trimmer. This means we install something that isn't used most of the time

Maarten-vd-Sande:
Also not sure how e.g. `gimmemotifs.coverage_table` works with those "unfiltered" bams

Maarten-vd-Sande:
@siebrenf I see that it's still in the multiqc yaml! :zipper_mouth_face:

Issue #664, opened 2021-03-15
FR: markduplicates and primary align defaults


Maarten-vd-Sande:
> repetitive reads (multi-mappers) should **not** be removed from the BAM. If that happens, that should change in the pipeline.
> While in general we discard them (for TF ChIP-seq) there are ways to use them, as is the case in Stefan's project.
> So they should be present in the final BAM (though marked as multi-mapper)
> This is for the BAMs though. Default setting for bigWigs should be to remove both multimappers as well as duplicates.
> The primary alignment (picked at random) will not have flag 256 set, but will have a mapq of 0

Easy enough to revert this (we changed to remove duplicates by default in 0.3.0). Not sure how we can circumvent the `mapq` problem (in that filtering for mapq seems nice)

Maarten-vd-Sande:
Macs2 documentation (is gone btw :shrug:) is completely unclear about which reads are considered for peak calling and which arent (this is relevant for the resulting bigwigs). Some old discussion: https://groups.google.com/g/macs-announcement/c/ASQ3GsXQC1c

Not looking forward to have a pre-sieve bam, post-sieve bam, and a peak-call bam...

Maarten-vd-Sande:
Also not sure how e.g. `gimmemotifs.coverage_table` works with those "unfiltered" bams

Maarten-vd-Sande:
Perhaps a single flag that sets it to keep everything (but marked) and also a FAQ entry? :thinking:

siebrenf:
not the way we want to implement this, but on the bigwig side its achieved with this:
```
  deeptools_flags:
    description: flags for bam to bigwig conversions with deeptools bamCoverage. Bins Per Million mapped reads is roughly the same as TPM in RNA-seq.
    default: '--normalizeUsing BPM --samFlagExclude 1024 --minMappingQuality 1'
```

Maarten-vd-Sande:
Yes, however the bigwigs are generated by macs2/genrich when you use those for peak calling....!

Maarten-vd-Sande:
@siebrenf perhaps it's possible to add the result of `seq2science explain` at the top of the multiqc report? That would be kinda nice I guess?

PR #665, opened 2021-03-16
Explain qc

Maarten-vd-Sande:
Related to #664 to make it more clear what we are doing. 

**todo**

- [ ] @siebrenf, is this approach okay/maintainable enough? I'm not sure how else, `_explain` isn't easier I think...
- [ ] Better explain tests, there were some errors introduced and tests passed :see_no_evil: 
- [ ] Not sure how easy/possible it is to make hyperlinks. Much nicer to read actually.
- [ ] Add explanation to FAQ what the final bam is

image

**Checklist**
- [ ] I updated the CHANGELOG


siebrenf:
The explain feature is neat, and this would be another way of bringing this feature info to the user. 
However, it is a wall of text. I'm not sure if it would be read more than the explain function itself this way.

What you could do is add the explain to the bottom of the multiQC, and add a link/verbal reference in it in the BAM QC section header.

Maarten-vd-Sande:
It does not only relate to the BAM stuff. So I am not sure about linking it in the bam qc section. Another thing is that I am afraid when we put it on the bottom, that people (again) won't read it. Maybe I can make hyperlinks instead of those urls in brackets. But not sure how that would translate on the terminal

Maarten-vd-Sande:
@siebrenf and if I were to make each url a hyperlink? No hyperlink here, but the text is much nicer to read already (and shorter)
image

siebrenf:
You could drop all references here, and refer to the explain function. Additionally, some text from the explain feature can be shortened (removing `with options ""` if there are none). Together the explain text might get small and readable enough :)

btw, explain tests are now working & expanded in the develop branch again

siebrenf:
its actually not that much work. Ill test it out!

Maarten-vd-Sande:
Cool, because I like what you proposed. Much more consistent

siebrenf (approved):
on the TODOs:
- the various explain methods should be easier to read without the regex. maintainability ✔️ 
- explain tests were fixed prior: ✔️ 
- sweet hyperlink functionality: ✔️ 
- final BAM is now sort-of explained in the sieve-bam rule: 🤷

Maarten-vd-Sande:
We would need to up the minimum python version for the walrus btw

PR #670, opened 2021-03-22
fix filtering issue

Maarten-vd-Sande:
Bug when filtering template length on a mix of single-end and paired-end reads

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
why have it in the output if it is
- not used anywhere
- touched in some cases

I'd move it to the params (if you can call the lambda with another param)

PR #674, opened 2021-03-25
DESeq2.R from CLI

siebrenf:
So you wanna perform DESeq2, but you dont want to run seq2science again? OK, now you can with 
```
Rscript /path/to/s2s/seq2science/scripts/deseq2.R --help
```

**What did change**
The DESeq2 script is now split into the core script and two wrappers: one for the rule, one for the user. 
The standalone script is a bit more limited: only 1 assembly, the "technical_replicates" column is always merged, and only default config options available...

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/scripts/deseq2.R
+++ b/seq2science/scripts/deseq2.R
@@ -144,12 +144,15 @@ cat('\n')
 expressed_genes <- as.data.frame(resLFC[order(resLFC$padj),])
 
 missing_genes <- rownames(counts)[!(rownames(counts) %in% rownames(expressed_genes))]
-unexpressed_genes <- as.data.frame(matrix(data = NA, ncol = ncol(expressed_genes), nrow = length(missing_genes)))
-rownames(unexpressed_genes) <- missing_genes
-colnames(unexpressed_genes) <- colnames(expressed_genes)
-unexpressed_genes[,'baseMean'] <- 0
-
-all_genes <- rbind(expressed_genes, unexpressed_genes)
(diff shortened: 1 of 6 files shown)


seq2science/scripts/deseq2.R
```diff
@@ -1,49 +1,18 @@
...
+  scripts_dir <- snakemake@params$scripts_dir
+  deseq_wrapper <- file.path(scripts_dir, "deseq2_rule.R")
+} else {
+  # parse command line input
+  cli_args    <- commandArgs(trailingOnly=F)
+  this_script <- sub("--file=", "", cli_args[grep("--file=", cli_args)])
+  scripts_dir <- dirname(this_script)
+  deseq_wrapper <- file.path(scripts_dir, "deseq2_standalone.R")
+}
+source(deseq_wrapper)
```
Maarten-vd-Sande, line 15: source executes the script?

Maarten-vd-Sande (approved):
Nice new feature. 

Implementation feels kinda strange, but splitting over files has a clear readability advantage.

We can make the script an executable, so calling it is easier? Something like `deseq2science` ^^. I think it's possible through pip, otherwise we can add it to the bioconda recipe. However, do we want those packages then to be as a default in the environment?

seq2science/scripts/deseq2.R
siebrenf: yea, and global variables carry over. R can be weird...

PR #675, opened 2021-03-26
Jos proofing

siebrenf:
Writing the config can be confusing. It would be nice to get some help/informative errors.

**What did change**
- all config keys are converted to lower case (we dont use any upper case, so this removes user capitalization typos, for instance when copy-pasting from a chat board which auto-capitalizes)
- throws an informative error when a duplicate key is found in the user config

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/configuration_generic.smk
+++ b/seq2science/rules/configuration_generic.smk
@@ -25,7 +25,7 @@ from snakemake.utils import min_version
 from snakemake.exceptions import TerminatedException
 
 import seq2science
-from seq2science.util import samples2metadata, prep_filelock, url_is_alive, color_parser
+from seq2science.util import UniqueKeyLoader, samples2metadata, prep_filelock, url_is_alive, color_parser
 
 
 # get the cache and config dirs
@@ -33,6 +33,8 @@ CACHE_DIR = os.path.join(xdg.XDG_CACHE_HOME, "seq2science", seq2science.__versio
 
 # config.yaml(s)
 
+# convert all config keys to lower case (upper case = user typo)
+config =  {k.lower(): v for k, v in config.items()}
 
 # check config file for correct directory names
 for key, value in config.items():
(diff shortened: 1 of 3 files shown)


seq2science/rules/configuration_generic.smk
```diff
@@ -25,14 +25,16 @@ from snakemake.utils import min_version
...
+from seq2science.util import UniqueKeyLoader, samples2metadata, prep_filelock, url_is_alive, color_parser
 
 
 # get the cache and config dirs
 CACHE_DIR = os.path.join(xdg.XDG_CACHE_HOME, "seq2science", seq2science.__version__)
 
 # config.yaml(s)
 
+# convert all config keys to lower case (upper case = user typo)
+config =  {k.lower(): v for k, v in config.items()}
```
Maarten-vd-Sande, line 37: Not a fan! Which one would be kept? Better to crash :fire: :see_no_evil:

seq2science/rules/configuration_generic.smk
```diff
@@ -66,7 +68,8 @@ assert sorted(config['fqext'])[0] == config['fqext1'], \
 
 # read the config.yaml (not the profile)
 with open(workflow.overwrite_configfiles[0], 'r') as stream:
-    user_config = yaml.safe_load(stream)
+    # safe_load() with exception on duplicate keys
+    user_config = yaml.load(stream, Loader=UniqueKeyLoader)
```
Maarten-vd-Sande, line 72: perhaps catch the error and pass it to the logger?

Maarten-vd-Sande (approved):

seq2science/rules/configuration_generic.smk
siebrenf: good one, will do.

seq2science/rules/configuration_generic.smk
siebrenf: that's what happens in the Loader!

seq2science/rules/configuration_generic.smk
siebrenf: The loader checks for this now. This line is still in, in case the lowered case kay is not overlapping existing keys.

JGASmits:
I am crying and laughing at the same time 
＼（〇_ｏ）／

JGASmits (approved):

PR #679, opened 2021-04-02
doc fixes

siebrenf:
Sneaky push to master to fix the docs only.

The missing _whiteline with an extra space_ in `bam_cleaning.smk` was breaking the chapter "per rule explanation"...
The rest just looks a little better.


Maarten-vd-Sande (approved):
This kinda stuff you can just push directly to master branch, no need for review (in my opinion). :+1:

PR #685, opened 2021-04-19
no support for ABI solid

### LINKED ISSUE #533: BUG: ABI solid not supported
**Describe the bug**
ABI solid sequencing does not play well with most aligners. 

Perhaps an error/warning would be nice

Maarten-vd-Sande:
Fixes #533, now we give a nice error :smile: 

One thing to consider is to change it into a warning, with a pause of a couple secs, or add a flag to ignore the error. Stuff obviously doesn't always work, so being able to ignore this might be nice? 
  
**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/util.py
+++ b/seq2science/util.py
@@ -150,6 +150,10 @@ def samples2metadata_sra(samples: List[str], logger) -> dict:
                      "seq2science does not support downloading those..\n\n")
         raise TerminatedException
 
+    # keep track of not-supported samples
+    not_supported_formats = ["ABI_SOLID"]
+    not_supported_samples = []
+
     for sample, clean in sample2clean.items():
         # table indices
         idxs = _sample_to_idxs(df_sra, clean)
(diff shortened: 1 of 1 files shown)


siebrenf (approved):
Compact solution, nice. 

The internet suggests only BWA (used to) work(s) with SOLiD*. If you really want to support an unsupported data type, you could force the use of BWA as aligner, rather than adding more flags.

*sources:
Bowtie1 supports it, but we don't support Bowtie1.
[Bowtie2 doesn't support it](https://www.biostars.org/p/123508/)
HISAT2 is by the same group, and came after Bowtie, so that won't support it.
[STAR doesn't support it](https://groups.google.com/g/rna-star/c/GHuuqFQNLjk)
[BWA might support it](https://www.biostars.org/p/31/)

Maarten-vd-Sande:
The reason I proposed a flag to ignore it, is because stuff isn't always properly annotated on the web. So in the special case this happens you can ignore this. But maybe we should not think about these things until they happen

PR #689, opened 2021-04-23
redirect stderr

siebrenf:


Maarten-vd-Sande (commented):
I don't see how this solves the problem. It should be redirected to the logger if anything. The problem is that the seq2science logfile does not contain the correct log output. If I understand the issue correctly not only is the exception nog in the logfile, but also the `error in rule ...` part.

siebrenf:
Both the [Python](https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler) and the [Snakemake](https://snakemake.readthedocs.io/en/stable/_modules/snakemake/logging.html#Logger.set_stream_handler) logging modules can only handle one stream. AFAIK there is no "universal" stream, so this merges stderr with stdout, as [stdout is send to the logger](https://github.com/vanheeringen-lab/seq2science/blob/1893d585231e7d99bd3f8e3bdcc2ba48ded1ac5c/seq2science/cli.py#L364).

Assumption on my part here is that `error in rule...` is also send to stderr.

Maarten-vd-Sande:
Oh wow, I already forgot how I implemented this :see_no_evil: 

Awesome, I will make sure to test if this works, but would be a really easy way to solve it indeed :muscle:  Thanks

PR #691, opened 2021-04-28
catch jsondecodeerror

Maarten-vd-Sande:
Catches the JSONDecodeError from UCSC not working properly, and log it. However genomepy now also started having troubles.... I think those should be solved on genomepy side? Not sure

**Checklist**
- [ ] I updated the CHANGELOG


tilschaef (approved):

Maarten-vd-Sande:
What happenend?

siebrenf:
I couldn't find an easy way to add to this PR after it was pushed, but I had to change the genomepy-related code to actually use genomepy's internal solution for this problem :)

PR #695, opened 2021-05-11
stop storing 'local' provider

siebrenf:
**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/configuration_generic.smk
+++ b/seq2science/rules/configuration_generic.smk
@@ -246,32 +246,39 @@ if "assembly" in samples:
                 if os.path.exists(providersfile):
                     providers = pickle.load(open(providersfile, "rb"))
 
-                if any([assembly not in providers for assembly in set(samples["assembly"])]):
-                    logger.info("Determining assembly providers")
+                # list assemblies not in the s2s cache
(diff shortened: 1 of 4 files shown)


seq2science/rules/configuration_generic.smk
```diff
@@ -246,32 +246,39 @@ if "assembly" in samples:
...
+                    # check if the genome can be downloaded
+                    if providers[assembly]["genome"] is None:
+                        genome_provider = provider_with_file("genome", assembly)
+                        if genome_provider:
+                            providers[assembly]["genome"] = genome_provider
+                        # if the genome cannot be downloaded, but exists locally, leave it be.
+                        elif os.path.exists(f"{file}.fa"):
+                            providers[assembly]["genome"] = "local"
 
                     pickle.dump(providers, open(providersfile, "wb"))
```
Maarten-vd-Sande, line 283: I guess we want to not include the "local" ones in the providersfile either?

siebrenf: you're right! this dict only gets accessed when downloading, so local assemblies can just be skipped.

Maarten-vd-Sande (approved):
sorry I thought I approved this before

PR #702, opened 2021-05-20
assembly-workaround

siebrenf:
something like this?

### DIFF
--- a/seq2science/rules/quantification.smk
+++ b/seq2science/rules/quantification.smk
@@ -267,32 +267,36 @@ elif config["quantifier"] == "kallistobus":
                  """        
         
     if config.get['mode'] == "kite":
+        ruleorder: kallistobus_ref_kite > get_genome
         
         rule kallistobus_ref_kite:
             """
             Make a mismatch index for kallistobus. This index is required to count feature barcodes, such as antibody tags. 
             """
             input:
-                featurebarcodes=expand("{genome_dir}/{kite_fm}.tsv", **config)
+                featurebarcodes=expand("{genome_dir}/{assembly}.tsv", **config)
             output:
-                directory(expand("{genome_dir}/{kite_fm}/index/kallistobus_kite", **config)),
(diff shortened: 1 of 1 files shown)


tilschaef:
That still triggers genomepy to search for an assembly.  Ideally, if the user selects kite as workflow, the following should happen:

- Use `kallistobus_re_kite` to obtain the "reference" fasta after converting from tsv to .fasta instead of get genome (I think that is solved in this PR)
-  Do not automatically use genomepy to obtain the primary assembly and resources since they do not exist for the artifical reference

Would it be a general idea to have this kind of functionallity in seq2science, basically using local reference resources exclusively?

tilschaef:
I did a quick check and disabling the checks in `configuration_generic.smk` from `287-305` worked in combination with the changes @siebrenf  proposed.  In this way,  `get_genome` can be "overruled"  and one can supply a custom reference

Maarten-vd-Sande:
I have **no clue** what this is about :see_no_evil: , however one thing to keep in mind is that we have spent **a lot** of effort to make this work as well as possible for any species. Single-cell RNA-seq is a mess, with barcodes and all that stuff, but if possible, this should be supported for other species than human alone.

Might not be even relevant though, since I dont know what kite is :upside_down_face:

tilschaef:
> 
> 
> I have **no clue** what this is about 🙈 , however one thing to keep in mind is that we have spent **a lot** of effort to make this work as well as possible for any species. Single-cell RNA-seq is a mess, with barcodes and all that stuff, but if possible, this should be supported for other species than human alone.
> 
> Might not be even relevant though, since I dont know what kite is 🙃

@Maarten-vd-Sande  [kite](https://github.com/pachterlab/kite) is basically an alternative workflow for kallisto to create an index from antibody tags rather than transcripts. The starting point is a .tsv file, if you use the kb-wrapper, which contains the antibody sequences and identifiers. In that regard, you don't have a species assembly for a particular species to begin with. Kallisto creates the fasta, indexes etc. from that .tsv file.

The idea to get this working in s2s would be to disable the genome/annotation checks mentioned above for that particular workflow.

Issue #701, opened 2021-05-20
FR: Improving seq2science explain


simonvh:
First of all, the `seq2science explain` option is awesome.  Saves a lot of time writing the methods section, really appreciated!

I do have some suggestions for improvement (ATAC-seq workflow):
* Change the writing style to be consistent: *we aligned* versus *were aligned*. Passive style is OK for methods, change everything to consistently use passive style.
* Change `with options ''` to `with default options`.\
* `And finally we made a count table from the conensus [...]` -> `Finally, a count table was made from the consensus [...]`
* Change http://broadinstitute.github.io/picard to paper reference
* `with 'config[slop]' ` -> change to value from the config.

Maarten-vd-Sande:
1. Done
2. Done
3. Done
4. There is no paper, see FAQ: http://broadinstitute.github.io/picard/faq.html
5. Done (was already fixed in newer release I believe)

This is now the result of the default atac-seq workflow:

```
[...]

PR #703, opened 2021-05-20
explain fixes

Maarten-vd-Sande:
Related to #701 

**Checklist**
- [ ] I updated the CHANGELOG


seq2science/rules/configuration_explain.smk
```diff
@@ -108,7 +108,7 @@ else:
             assert isinstance(ret,dict), f"Invalid config key '{opt}' in {opts}."
             ret = ret.get(opt)
             if ret in [None, ""]:
-                return ""
+                return " with default options"
```
Maarten-vd-Sande, line 111: @siebrenf would this work, or are there edge cases I don't think of? (and the tests miss those)

siebrenf (approved):

PR #708, opened 2021-05-31
refactor genomepy rules

siebrenf:
genomepy related rules are doing too much at once, giving snakemake a headache.

**What did change**
All genomepy rules have been refactored. Each file is now downloaded separately, and (mostly) through the front end API (I'll try to make the front end more amenable to this too in the future).

Most importantly, the blacklist is now downloaded separately. If genomepy cannot download a blacklist because it does not exist, the new rule outputs an empty file.

**Checklist**
- [ ] I updated the CHANGELOG


seq2science/scripts/get_genome_annotation.py
```diff
@@ -1,41 +1,50 @@
...
 
 
 with open(snakemake.log[0], "w") as log:
     with contextlib.redirect_stdout(log), contextlib.redirect_stderr(log):
-        try:
-            provider = snakemake.params.providers[snakemake.wildcards.raw_assembly]["annotation"]
+        # list user plugins
+        active_plugins = genomepy.functions.config.get("plugin", [])
+        # deactivate user plugins
+        genomepy.funcions.manage_plugins("disable", active_plugins)
```
Maarten-vd-Sande, line 14: This breaks when doing this in parallel, not that I really care tbf :innocent:

Maarten-vd-Sande: If we want to do this properly, this should come from Snakemake. And onexit and onerror these should be set back. Perhaps even easier to do the whole thing there?

Maarten-vd-Sande: Somewhere here, although that is super confusing this is in loggen :yum: 

https://github.com/vanheeringen-lab/seq2science/blob/develop/seq2science/rules/configuration_logging.smk#L7-L22

Maarten-vd-Sande: Anyways, I don't think it really matters

siebrenf: all genomepy rules share resource `parallel_downloads=1`, so there should be no parallel genomepy runs.

Maarten-vd-Sande: parallel_downloads is set to three though :upside_down_face:

Maarten-vd-Sande: (in that three downloads run in parallel0

Maarten-vd-Sande: This is to bypass the problem when downloading e.g. two samples, where one is five time the size of the other. This guarantees a more continuous stream of files being downloaded for downstream processing. Anyways, I don't think we really have to worry about these cases

siebrenf: its an edge case yea. I've added a genomepy downloads resource too. should be fixed forever now :)

PR #720, opened 2021-06-15
Optional barcode whitelist

tilschaef:
kb-python's barcode argument  is optional since it comes already pre-packed with several whitelists, such as for 10X, in case they are not provided by the user. Currently, the barcode argument is mandatory when running the scRNA seq workflow. 

**What did change**
The `barcodefile` argument is now optional.

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/quantification.smk
+++ b/seq2science/rules/quantification.smk
@@ -280,9 +280,9 @@ elif config["quantifier"] == "kallistobus":
             Align reads against a transcriptome (index) with kallistobus and output a quantification file per sample.
             """
             input:
-                barcodefile=config["barcodefile"],
                 basedir=get_kb_dir,
-                reads=rules.fastq_pair.output.reads
+                reads=rules.fastq_pair.output.reads,
+                barcodefile=expand(["{barcodefile}"] if 'barcodefile' in config else [],**config),
             output:
                 dir=directory(expand("{result_dir}/{quantifier}/{{assembly}}-{{sample}}",**config))
             log:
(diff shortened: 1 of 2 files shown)


seq2science/rules/quantification.smk
```diff
@@ -280,9 +280,9 @@ elif config["quantifier"] == "kallistobus":
             Align reads against a transcriptome (index) with kallistobus and output a quantification file per sample.
             """
             input:
-                barcodefile=config["barcodefile"],
                 basedir=get_kb_dir,
-                reads=rules.fastq_pair.output.reads
+                reads=rules.fastq_pair.output.reads,
+                barcodefile=expand(["{barcodefile}"] if 'barcodefile' in config else [],**config),
```
siebrenf, line 285: it's nice that you have this in the input fields, as it's clear in the DAG this file is used.

not sure if this works, but couldn't the file path simply be `config.get("barcodefile", []),`?

seq2science/rules/quantification.smk
```diff
@@ -298,14 +298,15 @@ elif config["quantifier"] == "kallistobus":
                 mem_gb=66,
             params:
                 basename=lambda wildcards, input: f"{input.basedir[0]}/{wildcards.assembly}",
+                barcode_arg=lambda wildcards, input: "-w " + input.barcodefile[0] if input.barcodefile else "",
```
siebrenf, line 301: again, not sure if I'm right, but if `False`, wont `"-w " + input.barcodefile[0] if input.barcodefile else ""` give `-w `?
if so, solution is  `("-w " + input.barcodefile[0]) if input.barcodefile else ""`

siebrenf (approved):

seq2science/rules/quantification.smk
tilschaef: yes, that one gives equivalent results but is quite a bit shorter 👍

seq2science/rules/quantification.smk
tilschaef: If that expression is true, it returns the complete part before the if and "" otherwise. At least the command line output did not contain an empty `-w`. However, putting it in braces makes it a bit more readable.

Issue #707, opened 2021-05-31
BUG: kb count failing but s2s does not register


Maarten-vd-Sande:
**Describe the bug**
kallistobus count fails, but does not give an error code. @Rebecza has logs of this. @Rebecza proposed to check if in the logfile some keyword is present, and if so, fail/succeed based on that. Should be easy to implement. Another option would be to actually fix kb-python, but that is more involved, and will only work once they release that fix.

tilschaef:
@Rebecza , can you attach the logfie to the issues? I also noticed that kb-python would sometimes "silently" fail and not produce any output files while the kb count rule would still succeed!

tilschaef:
I think it should be enough to check if the log file contains ERROR and then let the rule exit

Maarten-vd-Sande:
Sure anything is fine by me. Are you sure a succesful run does not get sth like error? And a failed run?

Rebecza:
> 
> 
> @Rebecza , could you attach your logfie to this issues? I also noticed that kb-python would sometimes "silently" fail and not produce any output files while the kb count rule would still succeed! This was usually the case when the barcode layout /technology was not recognized.

I finally found back an example! 

This is the kind of error one randomly gets sometimes (because of storage issue?), so far for this error I never found any other reason, than just a short blib on the server causing it, because after trying again it works for that plate as well.

Therefore for this type of error specifically, it would be good to check the logs after s2s is finished, and run again for the plates where it finds: `std::bad_alloc`.

An example log file for this: 

```
[2020-12-15 20:31:55,800]   DEBUG Printing verbose output
[2020-12-15 20:31:55,800]   DEBUG kallisto binary located at /vol/mbconda/snabel/anaconda3/envs/seq2science/lib/python3.7/site-packages/seq2science/.snakemake/dc8b83b3/lib/python3.8/site-packages/kb_python/bins/linux/kallisto/kallisto
[2020-12-15 20:31:55,800]   DEBUG bustools binary located at /vol/mbconda/snabel/anaconda3/envs/seq2science/lib/python3.7/site-packages/seq2science/.snakemake/dc8b83b3/lib/python3.8/site-packages/kb_python/bins/linux/bustools/bustools
[2020-12-15 20:31:55,800]   DEBUG Creating /scratch/snabel/scrna/data_d4-26_all_s2s/results/kallistobus/GRCh38-Mon-AM-m5-1-Z41_S3/tmp directory
[...]

PR #722, opened 2021-06-18
Added checks

tilschaef:
Addresses issues described in #707 

**What did change**

1. Check kb log file for ERROR/bad_alloc keyword and exit rule if encourtered
2. Delete corrupted output folder if 1.

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/quantification.smk
+++ b/seq2science/rules/quantification.smk
@@ -309,8 +309,6 @@ elif config["quantifier"] == "kallistobus":
                 {params.barcode_arg} {params.options} {input.reads} > {log} 2>&1
                 # Validate output
                 if grep -q 'ERROR\|bad_alloc' "{log}"; then
-                  echo "An error was encoutered. Please check the logfile: {log} for details!" 
-                  if [ -e "{output}" ]; then rm -Rf {output}; fi
                   exit 1
                 fi  
                 """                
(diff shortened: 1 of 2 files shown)


seq2science/rules/quantification.smk
```diff
@@ -307,6 +307,12 @@ elif config["quantifier"] == "kallistobus":
                 -t {threads} -g {params.basename}_t2g.txt \
                 -o {output} -c1 {params.basename}_cdna_t2c.txt -c2 {params.basename}_intron_t2c.txt \
                 {params.barcode_arg} {params.options} {input.reads} > {log} 2>&1
+                # Validate output
+                if grep -q 'ERROR\|bad_alloc' "{log}"; then
+                  echo "An error was encoutered. Please check the logfile: {log} for details!"
```
Maarten-vd-Sande, line 312: `encountered`

Maarten-vd-Sande: Also it's better not to print to shell because when running lots of stuff in parallel. Better to put some comment in the log, that we manually failed the rule (or no comment at all, I think the log is quite self-explanatory)

seq2science/rules/quantification.smk
```diff
@@ -307,6 +307,12 @@ elif config["quantifier"] == "kallistobus":
                 -t {threads} -g {params.basename}_t2g.txt \
                 -o {output} -c1 {params.basename}_cdna_t2c.txt -c2 {params.basename}_intron_t2c.txt \
                 {params.barcode_arg} {params.options} {input.reads} > {log} 2>&1
+                # Validate output
+                if grep -q 'ERROR\|bad_alloc' "{log}"; then
+                  echo "An error was encoutered. Please check the logfile: {log} for details!" 
+                  if [ -e "{output}" ]; then rm -Rf {output}; fi
+                  exit 1
```
Maarten-vd-Sande, line 314: I think you can just `exit 1`, snakemake will remove the file for you

tilschaef: Nice, didn't know that :)

PR #724, opened 2021-06-20
Genome size non-hacky :)

Maarten-vd-Sande:
Before hacky genome size. Now reusable, and sensible approach :)

**Checklist**
- [x] I updated the CHANGELOG


seq2science/rules/get_genome.smk
```diff
@@ -158,3 +158,28 @@ rule gene_id2name:
...
+    Get the effective genome size for a kmer length. Some tools (e.g. macs2) require
+    an estimation of the effective genome size to better estimate how (un)likely it
+    is to have a certain number of reads on a position. The actual genome size is
+    not the best indication in these cases, since reads in repetitive regions
+    (for a certain kmer length) can not possible align on some locations.
+    """
+    input:
+        expand("{genome_dir}/{{assembly}}/{{assembly}}.fa", **config),
+    output:
+        expand("{genome_dir}/{{assembly}}/{{assembly}}.kmer_{{kmer_size}}.genome_size", **config)
```
siebrenf, line 174: how many files should I expect? If its more than 5, maybe make a subdirectory?

seq2science/rules/qc.smk
```diff
@@ -116,11 +116,11 @@ if config["trimmer"] == "trimgalore":
             """
 
 
-elif config["trimmer"] == "fastp":
+elif config["trimmer"] == "fastp" or "kmer_estimation" in config:
```
siebrenf, line 119: what's with this `or`? We only support these 2 trimmer right?

seq2science/rules/qc.smk
```diff
@@ -116,11 +116,11 @@ if config["trimmer"] == "trimgalore":
...
     ruleorder: fastp_qc_PE> fastp_qc_SE > fastp_PE > fastp_SE
 
 
-    rule fastp_qc_SE:
+    checkpoint fastp_qc_SE:
```
siebrenf, line 123: I don't think this one needs to be a checkpoint. It isn't referenced anywhere at least.

seq2science/rules/configuration_workflows.smk
```diff
@@ -139,3 +140,47 @@ if config.get("contrasts"):
             f"\nCould not parse DESeq2 contrast '{contrast}'.\n"
             "A DESeq2 design contrast must be in the form '(batch+)column_target_reference'. See the docs for examples.\n")
         _,_,_,_ = parse_contrast(contrast, samples, check=True)
+
+
+def get_read_length(sample):
```
siebrenf, line 145: Do the inputs need to be checkpoints? Would it not be faster to add these qc files/`rules.get_effective_genome_size.output` as input for macs2? My concern is speed here. If that is not applicable, then nvm :)

siebrenf (commented):
In general, speeding up macs is nice! Not sure how fast/easy checkpoints are now though. Some of these don't look like they need to be checkpoints anyway!

seq2science/rules/configuration_workflows.smk
Maarten-vd-Sande: The read length is also used by e.g. the chromap aligner, which I'll implement now that we finally have working checkpoints. I don't think checkpoints have a big performance hindrance.

Also if we don't use checkpoints on estimation of the read lengths, we would have to estimate the genome size for each sample separately, which is the whole reason i did this :)

seq2science/rules/qc.smk
Maarten-vd-Sande: Oops :see_no_evil:

seq2science/rules/qc.smk
Maarten-vd-Sande: :+1:

siebrenf (approved):

PR #725, opened 2021-06-21
read_csv error feedback

siebrenf:
image

**What did change**
added a line explaining this was caused by the samples.tsv

**samples.tsv to reproduce the error:**
```
sample	assembly
GSM1483769	GRCz11
GSM1483770	GRCz11	a
```


Maarten-vd-Sande (commented):
I am not sure this is how it works. My understanding is that `e` is the traceback, and not the exception. Not sure what happens here if you `raise e`. 

Perhaps we should print the samples file, so that it (maybe) becomes clear how s2s sees the file?

siebrenf:
We can't print the samples, because the error occurs in `ps.read_csv()`. 

I've changed the feedback now:
- if the error of this type, response is:
    ```
    An error occurred while reading the samples.tsv:
    Expected 2 fields in line 3, saw 3
    ```
- else a generic warning is given that this is cause by the samples.tsv, followed by the full traceback.
   ```
   An error occurred while reading the samples.tsv:

   [traceback]
   ```

Maarten-vd-Sande (approved):
I did not know the actual Error class gets passed, somehow I thought it was a string!

What do you think about making the error something like this? Either way it is fine by me, much less scary to get this error than the mega traceback from before

```
column_error = re.match("Expected (\d+) fields in line (\d+), saw (\d+)", string)
logger.error(f"We found {column_error.group(1)} columns on line {column_error.group(2)}, but seq2science excepts only {column_error.group(3)} columns...")
```

siebrenf:
with the new-new feedback:
image

PR #731, opened 2021-07-02
dontbestupid rerunning

Maarten-vd-Sande:
rerunning can be stupid, in that if the order of input changes due to whatever reason, it thinks it has to redo the whole rule. This PR is here to fix that :muscle: 

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


Maarten-vd-Sande:
@siebrenf I think this fixes the annoying way rerunning is handled. Due to randomness, sometimes the order of files is different from other runs. This causes a trigger of the rerunning, which is not necessary. I did some suuuuuuuuuuuuuuuuper simple checks, and this **seems** to work properly now

siebrenf (approved):
oh yes PLEASE!

Issue #686, opened 2021-04-21
FR: markdups on pre-sieve


Maarten-vd-Sande:
**Is your feature request related to a problem? Please describe.**

Samtools stats reports, among many other things, the number of duplicates for a sample. However when duplicates have not been marked, this will be reported as **no duplicates**. Better to mark duplicates on the pre-sieve, so both samtools stats have the proper number of duplicates. This is especially relevant when removing duplicate reads

PR #733, opened 2021-07-12
Markdups first, so that stats of bam are correct

Maarten-vd-Sande:
See #686 

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/bam_cleaning.smk
+++ b/seq2science/rules/bam_cleaning.smk
@@ -127,6 +127,12 @@ if config["filter_on_size"]:
 else:
     sieve_bam_output = {"final": f"{sieve_dir}/{{assembly}}-{{sample}}.samtools-coordinate{shiftsieve}.bam"}
 
+sieve_flag = 0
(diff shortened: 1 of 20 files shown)


seq2science/rules/bam_cleaning.smk
```diff
@@ -124,6 +182,23 @@ rule sieve_bam:
         {params.sizesieve_touch}
         """
 
+if not sieve_bam(config):
+    rule cp_unsieved2sieved:
+        """
+        This simply copies a bam file before that is unsieved, to the final bam directory,
+        in case no sieving is necessary.
```
siebrenf, line 189: `Copy a bam file if no sieving is necessary. `

this is the worst I could do

siebrenf (approved):

seq2science/rules/bam_cleaning.smk
Maarten-vd-Sande: :see_no_evil:

PR #734, opened 2021-07-15
Broadpeaks figures fix

Maarten-vd-Sande:
Can't make the nice plots for broadpeaks 

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
I'll take your word for it

Maarten-vd-Sande:
The current implementation of broadpeaks is **very** limited. It can't make a count table for them, so all the count table related stuff also can't be done. It's something that would be nice to support, but I haven't needed it yet so it's not working :smile:

PR #739, opened 2021-08-04
Treps for scRNA, fixes #736

### LINKED ISSUE #736: FR: merge technical replica's scRNAseq pipeline
It would be nice to be able to merge fastqs of technical replica's in the scRNAseq pipeline similar to how the bulk pipelines support this. Right now this isnt working yet, if i inc. a column with 'technical replica'  it complains that the values in it should be unique.
No hurry though, CAT on the fastqs and inputting those in s2s seems to work fine!

Maarten-vd-Sande:
@tilschaef this is a start. I haven't tested anything except dryrun, and it seems to work so far. Can you check if this works? In the end I did not need ruleorders.. :see_no_evil: 

Also @tilschaef can you finish this PR in case it's not fully working? Also can you add a test to the dag to make sure that this doesn't get broken in the future?

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/merge_replicates.smk
+++ b/seq2science/rules/merge_replicates.smk
@@ -73,18 +73,20 @@ def rep_to_descriptive(rep, brep=False):
 if "technical_replicates" in samples:
 
     def get_merge_replicates(wildcards):
-        input_files = expand(
-            [
-                f"{{trimmed_dir}}/{sample}{wildcards.fqext}_trimmed.{{fqsuffix}}.gz"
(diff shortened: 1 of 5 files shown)


siebrenf (approved):
The changes I see look good! 

It won't break anything outside scRNA. Are there any QC files being generated per file? If so, check the treps dataframe.

Maarten-vd-Sande:
Yes there are, @tilschaef do you have a QC report? I can quickly check if the trimming qc works properly :smile:

PR #740, opened 2021-08-05
upgrade genomepy to 0.10

siebrenf:
**What's new?**
genomepy 0.10 is both faster and more robust.


seq2science/util.py
```diff
@@ -589,3 +590,63 @@ def save(self):
...
+        """
+        Check the genomepy database for a provider stat serves both genome and annotation for an assembly.
+        If impossible, settle with the first provider that serves the genome.
+        """
+        logger.info("Determining assembly providers")
+        for p in genomepy.providers.online_providers():
+            for assembly in search_assemblies:
+                if assembly not in self:
+                    self[assembly] = {"genome": None, "annotation": None}
+
```
Maarten-vd-Sande, line 604: Should we call this `search_assemblies`?

Maarten-vd-Sande (approved):

Maarten-vd-Sande:
Does the stuff about downloading the refseq stuff etc get into the seq2science log? E.g.

```
refseq: 167k genomes [00:01, 159k genomes/s]
refseq: 183k genomes [00:01, 150k genomes/s]
refseq: 198k genomes [00:01, 152k genomes/s]
refseq: 214k genomes [00:01, 154k genomes/s]
refseq: 230k genomes [00:01, 155k genomes/s]
refseq: 239k genomes [00:01, 143k genomes/s]
Could not download assembly Bl71nemr.
Find alternative assemblies with `genomepy search Bl71nemr`
```

I think that would be best if that happens, but if difficult, then ignore it

siebrenf:
all genomepy logs are now properly sinked. however, the tests are extremely flakey... 

commits "cleanup" and "open logger.logfile first?" had working docs. "add genomepy messages to s2s log" has jenkins running (I canceled a few after that).
It seems that seq2science getting the genomes in the config is somehow less stable when genomepy also needs to get its genomes still...

PR #749, opened 2021-09-03
Update bigfiles.smk

siebrenf:
image

This gene is located on the forward/plus strand. I suppose we could have a discussion on the semantics of the mRNA being reverse or not, but its easier to interpret when it matches the gene's strand designation (forward/plus).

**Checklist**

### DIFF
--- a/seq2science/rules/bigfiles.smk
+++ b/seq2science/rules/bigfiles.smk
@@ -136,9 +136,9 @@ rule peak_bigpeak:
 
 def set_strand(wildcards):
     if wildcards.strand == ".fwd":
-        return "--filterRNAstrand forward"
+        return "--filterRNAstrand reverse"  # drop reverse/minus/antisense strand reads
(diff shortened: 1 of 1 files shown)


Maarten-vd-Sande (approved):
Any clue what causes the failed CI?

Why are we loading providers, but then never update it? Could that be related?
https://github.com/vanheeringen-lab/seq2science/blob/develop/seq2science/rules/configuration_generic.smk#L221

siebrenf:
the "tinydata" genome shouldnt need a lookup, as all files are local. the error must have occured [in the check if all files exist locally](https://github.com/vanheeringen-lab/seq2science/blob/b19bd7d0f7c6936d7ca43fa7bbc2abfedfd2f0b8/seq2science/rules/configuration_generic.smk#L226). If it keeps up I'll have a look, but I guess it was a concurrency issue on jenkins.

PR #751, opened 2021-09-20
tweak regex

siebrenf:
(\/.+){2} = .../genome/genome...
.* = _custom suffix
rest = actual suffix

**Checklist**

### DIFF
--- a/seq2science/cli.py
+++ b/seq2science/cli.py
@@ -346,8 +346,8 @@ def _run(args, base_dir, workflows_dir, config_path):
         #      not want to consider.
         #      - genome files since provider will change to local
         regex_patterns = [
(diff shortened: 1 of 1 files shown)


Maarten-vd-Sande (approved):
Makes sense, did you test it? I'm pretty sure rerunning isn't CI tested

siebrenf:
Tested with the files we went over yesterday.  `(\/.+){2}.+` -> `(\/.+){2}.*` fixes the issue where the genome name was truncated (because `.+` captured at least 1 character).

The regex now captures:
/path/to/genome/genome*.fa*
/path/to/genome/genome*.gaps*
/path/to/genome/genome*.annotation.bed*
/path/to/genome/genome*.annotation.gtf*

the only off-target in this would be 
`/path/to/genome/genome.transcripts.fa`
but I don't see when you'd want to rerun that either.

PR #771, opened 2021-11-15
fix local samples edge case

Maarten-vd-Sande:
I had local samples (not publicly), thus they were in my cache. I removed the full results dir, and I got this weird error. Local samples shouldnt be stored in the cache I think.

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
makes sense!

PR #772, opened 2021-11-15
Fix assembly stats 

Maarten-vd-Sande:
DRAFT

**Checklist**
- [ ] I updated the CHANGELOG


siebrenf (approved):

Maarten-vd-Sande:
@siebrenf I wasn't sure whether to call it `genome_dir` or `genomeS_dir`.. Any preference? s2s uses genome_dir, but genomepy genomes_dir :upside_down_face:

siebrenf:
yea, I use both in genomepy. The `genomeS_dir` is the one where are _all_ the `genomeS` are kept, so I'd use that one if possible.

PR #776, opened 2021-11-22
Log

Maarten-vd-Sande:
When an error is raised, the traceback is very confusing and hides the "real" error message we often supply. We can just log an error and exit quietly :smile: 

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
That's lot of work! nice!

PR #782, opened 2021-11-29
fix standalone

siebrenf:
fix the DESeq2 standalone script


Maarten-vd-Sande (approved):
Does it make sense to add tests for this? Hopefully we just never have to touch this code anyways :innocent:

siebrenf:
We'd need (tiny) test data with at least 4 samples with DE genes. This is still a todo for the regular tests too!
I might be able to cheat this in tho... but not for this PR

PR #785, opened 2021-11-30
kb ref fixes

tilschaef:
- Fixed bug causing kb ref to crash when building non-velocity index.
- Updated kb version

**What did change**
- Added optional parameters to kb ref
- Removed obsolete input function

**Checklist**
- [X] I made a PR to develop (not master)
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/quantification.smk
+++ b/seq2science/rules/quantification.smk
@@ -299,9 +299,9 @@ elif  "scrna_seq" == get_workflow():
         def get_kb_dir(wildcards):
             #Get the correct assembly directory for ADT/Genome based workflows
             if 'kite' in config.get('ref',""):
-                return directory(expand("{genome_dir}/{{assembly}}/index/kallistobus/kite/", **config))     
+                return directory(expand("{genome_dir}/{{assembly}}/index/kallistobus/kite", **config))     
             else:
-                return directory(expand("{genome_dir}/{{assembly}}/index/kallistobus/", **config))
+                return directory(expand("{genome_dir}/{{assembly}}/index/kallistobus", **config))
                 
                             
         rule kallistobus_count:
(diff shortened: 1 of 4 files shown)


seq2science/rules/quantification.smk
```diff
@@ -272,41 +275,33 @@ elif  "scrna_seq" == get_workflow():
             input:
                 featurebarcodes=expand("{genome_dir}/{{assembly}}.tsv", **config)
             output:
-                directory(expand("{genome_dir}/{{assembly}}/index/kallistobus/kite/", **config)),
+                directory(expand("{genome_dir}/{{assembly}}/index/kallistobus", **config)),
```
siebrenf, line 278: If I see this correctly, the `kallistobus_ref` index and the `kallistobus_ref_kite` index now share the same output directory. Furthermore, the downstream `kallistobus_count` uses this directory, no matter the content.

If you were to change the index type, snakemake has no way to see which index is in the directory. Is this intended, or did I miss something?

tilschaef: I think you might be right, if someone would change the index type Snakemake would not be able to distinguish. I think that was the reason why I added it in the first place :)

siebrenf (approved):

seq2science/workflows/scrna_seq/config.yaml
```diff
@@ -34,3 +34,12 @@ quantifier:
...
 barcodefile: "barcodes.txt"
+
+#Export Seurat object
+export_seu_objects: True
+
+seurat_object:
+    project_name: merged
+    min_cells: 0
+    min_features: 0
+
```
Maarten-vd-Sande, line 45: This hasn't been added to the schema! :innocent:

PR #788, opened 2021-12-01
Added check for merged samples

tilschaef:
Fixed bug causing read_kb_counts.R script to stop when technical replicates are merged.

**Checklist**
- [x] I updated the CHANGELOG


seq2science/scripts/seurat/read_kb_counts.R
```diff
@@ -41,9 +41,14 @@ cat('\n')
 
 # Read cell meta data from samplesheet
 prep_cell_meta <- function(seu, sample_sheet) {
-  blacklist <- c("descriptive_name","technical_replicates")
+  blacklist <- c("descriptive_name")
   meta <- colnames(sample_sheet[setdiff(names(sample_sheet),blacklist)])
-  sample.meta <- sample_sheet[sample_sheet$sample==seu@project.name,]
+  #Lookup samples based on sample name/merged replicate name
+  if ("technical_replicates" %in% colnames(sample_sheet)) {
```
siebrenf, line 47: It's even more complex than this (because we're switching from python to R):
In s2s, we load the samples.tsv into a df, fill in blank rows for every present column, then remove all columns unless used.

To replicate this in R, I check for the `technical_replicates` column name in the df in [the rule params](https://github.com/vanheeringen-lab/seq2science/blob/c7f1d467ce9d2a86bb3984449d2fc80ea94d3a3e/seq2science/rules/deseq2.smk#L83), then pass this as a boolean to the R script. Once in R, I use [this function](https://github.com/vanheeringen-lab/seq2science/blob/c7f1d467ce9d2a86bb3984449d2fc80ea94d3a3e/seq2science/scripts/deseq2/utils.R#L34) to replicate the s2s samples df with the boolean variable.

This solution is pretty hacky. Ideally we'd transfer the Python dataframe to an R dataframe directly. If you feel like making one, I'd use it too!

Should this function be useful: you can load it into R using this code:
```
scripts_dir  <- snakemake@params$scripts_dir
deseq_utils <- file.path(scripts_dir, "utils.R")
source(deseq_utils)
```
where `scripts_dir` is another rule param.

Maarten-vd-Sande (approved):

siebrenf (changes requested):

seq2science/scripts/seurat/read_kb_counts.R
siebrenf: Let me know if I overlooked something, or if this comment makes no sense to you.
This comment refers to the [general config](https://github.com/vanheeringen-lab/seq2science/blob/c7f1d467ce9d2a86bb3984449d2fc80ea94d3a3e/seq2science/rules/configuration_generic.smk#L141), so I assumed its also applicable in scRNA-seq

Maarten-vd-Sande: Couldn't we just pass the dataframe as a string to R (params: samples.to_string()) and then read it in the R code (read.csv(text=snakemake.params))?

Less :bug: sensitive? Although I dont know if this works

siebrenf: if that's possible it would be great

tilschaef: It is indeed a bit more complex than expected, thanks for the function for parsing the samplesheet though, that's super useful. For now, I will stick to reading the samples.tsv via the config path and see if I can improve it later on. @Maarten-vd-Sande suggestion would be something to try out later on.

Maarten-vd-Sande: I made #790 for this

siebrenf (approved):

PR #789, opened 2021-12-01
snakemake-mamba works again

siebrenf:
mamba 0.19.0 appears to work with snakemake again!


Maarten-vd-Sande (approved):
ait

PR #791, opened 2021-12-02
multiqc in docs

Maarten-vd-Sande:
Adds an example multiqc report to the docs!

I added this report:
http://ocimum.science.ru.nl/multiqc_GRCm38.p6.html

Idk if that one is good? We could do sth different? With deseq? 

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ jobs:
           $CONDA/bin/conda config --add channels bioconda
           $CONDA/bin/conda install mamba
           $CONDA/bin/mamba env update --file requirements.yaml --name base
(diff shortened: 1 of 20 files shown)


siebrenf (approved):
ATAC-/ChIP-seq makes sense, the only addition to RNA-seq is DupRadar, which has internal examples.

Having an example is a nice, so this would totally do. If you wanna be super flashy, you could make a MultiQC with:
- a DESeq2 contrast
- reduced number of samples (for visual clarity)
- at least two of which are technical replicates that get merged (to show you can)
- add one or two samples from another assembly (to show you can)

Maarten-vd-Sande:
I added a deseq2 contrast. Too lazy for the others :innocent:

PR #792, opened 2021-12-02
use tmpdir!

Maarten-vd-Sande:

Use tmpdir instead of outdir for tmp files with samtools sorting. 

**Checklist**
- [ ] I updated the CHANGELOG


siebrenf (approved):
oh man. the `temp()` and `trap` could be removed now. Its fine to keep them, but its interesting to see how much simpler this makes everything!

Maarten-vd-Sande:
The trap and rm are there in case the program exits prematurely. All the `temp("file")` in theory could be replaced. Not sure if we want to go there :smile:

PR #794, opened 2021-12-03
Subsample

Maarten-vd-Sande:
Pitfall 3 of https://www.nature.com/articles/s41576-021-00434-9! Making sure that each sample contains the same number of reads. It's off by default, can be turned on by adding subsample to the config

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/schemas/config/alignment_general.schema.yaml
+++ b/seq2science/schemas/config/alignment_general.schema.yaml
@@ -216,3 +216,8 @@ properties:
   max_template_length:
     description: the maximum template length for paired-end reads
     type: integer
+
+  subsample:
+    description: the maximum amount of reads allowed in the final bam file.
+    type: integer
+    default: -1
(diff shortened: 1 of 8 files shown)


seq2science/schemas/config/alignment_general.schema.yaml
```diff
@@ -216,3 +216,8 @@ properties:
   max_template_length:
     description: the maximum template length for paired-end reads
     type: integer
+
+  subsample:
+    description: the maximum amount of reads allowed in the final bam file.
+    type: integer
+    default: -1
```
siebrenf, line 223: 🙈

siebrenf (commented):
This implementation does subsampling on bams greater than _n_, but ignores bams smaller than _n_. Isn't [sampling with replacement](https://www.biostars.org/p/319616/) the better option here?

siebrenf:
https://stats.stackexchange.com/questions/545180/is-sampling-with-replacement-better-than-sampling-without-replacement

siebrenf (approved):

PR #795, opened 2021-12-03
coverage table fixes

siebrenf:
The gimme coverage table command has several defaults that we have as config options, but didn't pass along.

**Checklist**
- [x] I updated the CHANGELOG


siebrenf:
The way ANANSE loads bam files is pretty much the same as `coverage table`. Which means that with the right settings, we could use this table directly in ANANSE.

This PR makes sure the settings match by default, leading the way for further s2s-ANANSE integration!

Maarten-vd-Sande (changes requested):
Lots of good changes (honestly that was a mess!) :see_no_evil: 

However I realize from the schema description that the difference between slop and peak_windowsize was very unclear (incorrect even). The **slop** decides the final width of peaks. The **windowsize** decides the threshold for when peaks between replicates should be considered the same/different. This is done based on the summit, if the summits of peaks are within distance windowsize, then they get merged. If not, they get are treated as independent peaks. Those peaks then all become the same width (based on slop). I think it is important these two are independent params. Since peaks can be wide (e.g. 2000) but with multiple independent peaks inside. However that also means that we potentially measure the same location multiple times... (I don't think there is a perfect solution though)

About default peak widths. The 200 is based on ATAC/narrow, and this is probably not always the correct width. However, 2000 is really wide as a default! I think too wide. Maybe we should add some explanation/comment in the config.yaml on how to use them, and when using e.g. ANANSE to set to 1000?

siebrenf:
As I'm not very familiar with the ChIP-seq standards I trust your judgement :)
I've added the details to [the ANANSE docs](https://anansepy.readthedocs.io/en/develop/input_data/#enhancer-activity).

Maarten-vd-Sande (approved):
nice. Can't wait to have some smooth s2s->ananse

PR #797, opened 2021-12-06
snakefmt applied

siebrenf:
I've applied snakefmt. There was a lot of formatting, mostly whitespace related.
As there were some issues with the tool, lets see if the tests can still parse everything first.


Maarten-vd-Sande (approved):
I'm not reviewing this :innocent:

PR #798, opened 2021-12-07
proposed changes

Maarten-vd-Sande:
I propose this :smile: (note that this is a PR to `coverage_table` branch)


Maarten-vd-Sande:
I'm not so sure about making the chip-seq peaks so wide. Because when people expect wide peaks, they generally use the --broad flag on macs2, so no count table is made anyways.

PR #804, opened 2021-12-13
libcrypto fix

tdewijs:
An issue with running seq2science, specifically in package infer_strandedness, error: "ImportError: libcrypto.so.1.0.0: cannot open shared object file: No such file or directory". 

**What did change**
We added a line in file gene_counts.yaml, where a new, specific version of openssl was installed. We tested it, and it worked. 

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/envs/gene_counts.yaml
+++ b/seq2science/envs/gene_counts.yaml
@@ -7,3 +7,4 @@ dependencies:
   - bioconda::htseq=0.12.4
   - bioconda::subread=2.0.1
   - bioconda::rseqc=4.0.0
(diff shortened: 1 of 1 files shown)


Maarten-vd-Sande (approved):
Thanks for the suggested changes. Looks good to me!

Is there a reason to use this specific version and build? There seem to be newer version. Does it work with `conda-forge::openssl=3.0.0`?

siebrenf:
We need to either upgrade samtools (or pysam?), or downgrade openssl. Either way the result is a functioning openssl package. Testing takes ages on the server with the error (due to our conda issues), and this method works, so we'll keep it for now.

PR #806, opened 2021-12-15
Mods

siebrenf:
making some functions more modular (for integration with ANANSE again)

oh, and I updated the check for the minimal snakemake version, since snakemake now checks for `snakefile_main`.

**Checklist**

### DIFF
--- a/seq2science/rules/configuration_generic.smk
+++ b/seq2science/rules/configuration_generic.smk
@@ -30,9 +30,9 @@ from seq2science.util import (
     samples2metadata,
     prep_filelock,
     url_is_alive,
-    color_parser,
     PickleDict,
     is_local,
+    get_contrasts,
 )
 
 
@@ -300,6 +300,8 @@ if "assembly" in samples:
             else (assembly + config["custom_assembly_suffix"])
         )
 
+    # list of DESeq2 output files
+    DE_contrasts = get_contrasts(samples, config, all_assemblies)
 
 else:
     modified = False
--- a/seq2science/rules/configuration_generic.smk
+++ b/seq2science/rules/configuration_generic.smk
@@ -21,12 +21,12 @@ from pandas_schema import Column, Schema
 from pandas_schema.validation import MatchesPatternValidation, IsDistinctValidation
 
 from snakemake.logging import logger
-from snakemake.utils import validate
-from snakemake.utils import min_version
+from snakemake.utils import validate, min_version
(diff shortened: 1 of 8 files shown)


seq2science/rules/configuration_generic.smk
```diff
@@ -508,7 +474,7 @@ wildcard_constraints:
 
 
 # make sure the snakemake version corresponds to version in environment
-min_version("5.18")
+min_version("6.12.3")
```
Maarten-vd-Sande, line 477: good one!

seq2science/rules/qc.smk
```diff
@@ -751,7 +751,7 @@ rule multiqc_schema:
         order = -954
         deseq2_imgs = ""
         deseq2_order = ""
-        for contrast in expand_contrasts(samples, config):
+        for contrast in expand_contrasts(samples, config["contrasts"]):
```
Maarten-vd-Sande, line 754: where is contrasts being set? Do you mean `DE_contrasts`?

seq2science/rules/qc.smk
```diff
@@ -854,10 +854,10 @@ def get_qc_files(wildcards):
         qc["files"].update(files)
 
     # DESeq2 all contrast plots
-    if "get_contrasts" in globals():
+    if "DE_contrasts" in globals():
```
Maarten-vd-Sande, line 857: should we maybe make `DE_contrasts` just an empty list as default? Not sure what is better here

Maarten-vd-Sande: Although maybe checking if it exists is the easier solution indeed

Maarten-vd-Sande (approved):
Nice refactoring, makes it much cleaner!

seq2science/rules/qc.smk
siebrenf: `DE_contrasts` contains the file paths to each DESeq2 output.
`expand_contrasts()` returns the actual contrasts, which are parsed from the config.

so this line is correct, however, we could also make it into another global list, or run `config["contrasts"] = expand_contrasts(samples, config["contrasts"])` somewhere in the config. Do you have a preference?

seq2science/rules/qc.smk
siebrenf: using a global variable or checking if a variable exists. seems like chosing the lesser evil :p

PR #807, opened 2021-12-22
scale deeptools with n samples

siebrenf:
with a large number of samples, this plot is very artistic, yet slightly less useful...
image

**What did change**
the size of the plot now scales with the number of bams. The new plot may not be any more readable in the MultiQC itself, but if the pixel density remains the same, you can now open it in a separate window and zoom in (similar to the DESeq2 heatmaps).

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/qc.smk
+++ b/seq2science/rules/qc.smk
@@ -494,12 +494,25 @@ rule multiBamSummary:
         """
 
 
+def get_plotCor_opts(lst):
+    opts = config["deeptools_plotcorrelation"]
+    n = len(lst)
+    if "--plotHeight" not in opts:
+        # default: 9.5
+        opts += f"--plotHeight {min(9.5, n*0.2}"
+    if "--plotWidth" not in opts:
+        # default: 11
+        opts += f"--plotWidth {min(11, 1.5+n*0.2}"
+    return opts
+
+
 rule plotCorrelation:
     """
     Calculate the correlation between bams with deepTools.
     """
     input:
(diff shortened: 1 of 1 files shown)


siebrenf:
The new figures can get too small to read in the MultiQC, but you can open them separately. 
P.S. the whitespace and DPI cannot be fixed.

0 30 GRCz11-deepTools_spearman_correlation_clustering_mqc

seq2science/rules/qc.smk
```diff
@@ -494,12 +494,25 @@ rule multiBamSummary:
...
+
+
 rule plotCorrelation:
     """
     Calculate the correlation between bams with deepTools.
     """
     input:
-        rules.multiBamSummary.output,
+        unpack(fingerprint_multiBamSummary_input),
+        cordata=rules.multiBamSummary.output,
```
Maarten-vd-Sande, line 515: Is this necessary?

Maarten-vd-Sande (approved):
Awesome!

seq2science/rules/qc.smk
siebrenf: `fingerprint_multiBamSummary_input` contains a list of bam files, which is used in `get_plotCor_opts()` to determine the number of samples. 

This seemed like the easiest source to me. If you have a better idea I can change it!

Maarten-vd-Sande: ah like that. Sure makes sense! :+1:

PR #809, opened 2022-01-13
Update edger.yaml

Maarten-vd-Sande:
Based on issue by @tdewijs, and inspired by https://github.com/bioconda/bioconda-recipes/issues/5350 I removed the R channel and requirements and all seems good.

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+### Fixed
+
+- corrupt environment creation with libreadline for edgeR normalization.
+
 ## [0.6.1] - 2021-12-17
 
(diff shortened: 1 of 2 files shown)


siebrenf:
I've seen similar errors in other R environments 😭. 

By unpinning the R version reproducibility goes down the toilet, but clearly some packages aren't pinned to begin with! Perhaps this could be fixed with dependency `conda-forge::r-base=3.6.1` and/or `conda-forge::readline=6.2`. I'll have a look.

siebrenf:
~Right. Ok. So inside the dependency hell that are R packages, there was a package with a broken link to lineread.so.~ Whichever package that might have been, it's been fixed now.

As for this PR: dropping the outdated R channel should improve stability.

siebrenf (approved):

PR #810, opened 2022-01-14
upsetplot

Maarten-vd-Sande:

When doing peak calling there is no real sense what the overlap/differences are between timepoints/samples. An upset plot is a easy way to get a feeling for these distributions. This is for instance an example of real data of a time series. It shows clearly that timepoint 1 has a lot of unique peaks compared to the other timepoints. It also shows that it is much more likely that timepoints that are next to each other have overlapping peaks than timepoints that aren't next to each other (duhhh!).

image

I also moved all peak-related QC to a peak qc snakefile

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (commented):
Great ideas all around. The only issue I have is the size of this plot: add 1 more sample and you'd more than double the number of columns. Can the plot be limited to the top 31 permutations?

Maarten-vd-Sande:
Makes sense! Currently not easily supported though :cry: :

https://github.com/jnothman/UpSetPlot/issues/173

Maarten-vd-Sande:
@siebrenf I think this is now implemented

siebrenf (approved):

PR #811, opened 2022-01-19
downgrade snakemake.. Yay....

Maarten-vd-Sande:

buggy new snakemake versions are a joy.. Snakemake revert/downgrade

**Checklist**
- [x] I updated the CHANGELOG


seq2science/rules/configuration_workflows.smk
```diff
@@ -153,47 +152,3 @@ if config.get("contrasts"):
...
-        qc_report = zipfile.ZipFile(qc_file, "r").read(zip_name)
-
-        kmer_size = re.search("Sequence length\\t(\d+)", qc_report.decode("utf-8")).group(1)
-
-    # fastp
-    if config["trimmer"] == "fastp":
-        if sampledict[sample]["layout"] == "SINGLE":
-            qc_file = checkpoints.fastp_SE.get(sample=sample).output.qc_json[0]
-        if sampledict[sample]["layout"] == "PAIRED":
-            qc_file = checkpoints.fastp_PE.get(sample=sample).output.qc_json[0]
```
siebrenf, line 191: It hurts to see you delete all your hard work :<

Suggestion: 
- turn this into a rule that outputs a temp file for each sample. 
- `macs2_callpeak` accepts this temp file as input, instead of the kmer size-specific file.
- the rule body executes the shell script in `get_effective_genome_size`, creating the kmer size-specific output, if it does not yet exist.
- the kmer size-specific file is copied and renamed to the temp output file.

that should remove the need for checkpoints, but now only 1 of these can run simultaneously to prevent double kmer file generation (and old/corrupted kmer files are never overwritten).

seq2science/rules/configuration_generic.smk
```diff
@@ -474,7 +476,7 @@ wildcard_constraints:
 
 
 # make sure the snakemake version corresponds to version in environment
-min_version("6.12.3")
```
siebrenf, line 477: since we'd need an exact version again:

```
import snakemake
from snakemake.exceptions import WorkflowError

if snakemake.__version__ != "5.18.1":
  raise WorkflowError(
    f"Expecting Snakemake version 5.18.1 (you are currently using {snakemake.__version__}
  )
```

siebrenf (changes requested):
I think we can salvage your code, if you think its worth it.

seq2science/rules/configuration_workflows.smk
Maarten-vd-Sande: I changed it so that there is now a genome size per sample, not per kmer size. Might decrease the down time a little bit.. :shrug:

siebrenf (approved):
troubleshooting takes longer than the extra rules ;)

PR #818, opened 2022-02-10
added pytxi

siebrenf:
Added pytxi to convert TPM to gene counts. Pytxi returns gene symbols btw.
Counts table use descriptive>replicate>sample names

todo: 
- ATAC-/ChIP-counts tables need to use descriptive names too
- clustering is failing for ATAC-seq peak counts

**Checklist**
- [x] I updated the CHANGELOG


seq2science/envs/pytxi.yaml
```diff
@@ -0,0 +1,10 @@
+name: pytxi
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - genomepy=0.11.1
+  - pip
+  - pip:
+    - git+https://github.com/vanheeringen-lab/pytxi.git@develop
```
Maarten-vd-Sande, line 10: Not sure if this is a permanent solution, but I think we should at least use a tagged version or sth like that..

siebrenf: yea sure! I'm running into some issue during my test run. Once those are fixed it use a commit hash/version tag!

seq2science/rules/peak_counts.smk
```diff
@@ -151,6 +151,13 @@ def get_coverage_table_replicates(file_ext):
 
     return wrapped
 
+def get_names(wildcards):
+    names = [""]
+    for rep in treps[treps["assembly"] == ori_assembly(wildcards.assembly)].index:
+        names.append(rep_to_descriptive(rep, brep=False))
+    names = "\t".join(names)
+    return names
+
```
Maarten-vd-Sande, line 160: I double checked, this _should_ be fine. I'm not happy with the way I set this up originally to begin with. Feels like it is **very** easy to make a mistake here... :eyes:

seq2science/rules/quantification.smk
```diff
@@ -31,7 +30,6 @@ if config["quantifier"] == "salmon":
         shell:
             "gffread -w {output} -g {input.fa} {input.gtf} >> {log} 2>&1"
 
-
```
Maarten-vd-Sande, line 34: :eyes:

seq2science/scripts/pytxi.py
```diff
@@ -0,0 +1,51 @@
...
+fnames = snakemake.input.cts
+assembly = snakemake.wildcards.assembly
+species = snakemake.input.fa[0]
+from_gtf = snakemake.params.from_gtf
+output = snakemake.output[0]
+sample_names = snakemake.params.names
+
+# redirect all messages to a logfile
+open(logfile, "w")  # start a clean log
+with open(logfile, "a") as log:  # appending because we mix streams (tqdm, logger & std)
```
Maarten-vd-Sande, line 22: Can't you just do `with open(logfile, "w") as log`? Not sure what happens if you open the file with 'w' and not close it

Maarten-vd-Sande (approved):

seq2science/rules/peak_counts.smk
siebrenf: its the same for loop used to gather the files, so it should always be the correct order. I apply the same method to convert the RNA counts to descriptive elsewhere.

I agree its hacky, so we could leave this out this PR and come back to it later?

seq2science/rules/quantification.smk
siebrenf: copy pasting error i guess.

seq2science/scripts/pytxi.py
siebrenf: nope, that caused issues for some reason. Loguru should be outputting to stderr, but its clearly more complex.

PR #819, opened 2022-02-15
remove the checkpoint

siebrenf:
checkpoints are still buggy. there was one, and it was still causing weird reruns. So I'm using bad bad practices to circumvent is. oh well...

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande:
I took the liberty of changing the default track resolution. Idk if that causes a hit on performance or anything?

seq2science/rules/configuration_explain.smk
```diff
@@ -145,7 +145,7 @@ else:
             end=" and finally were tn5 bias shifted by seq2science." if config.get("tn5_shift", 0) > 0 else ".",
         )
         + f"Afterwards samples were downsampled to {config['subsample']} reads."
-        if config.get("subsample")
+        if config.get("subsample", -1) != -1
```
Maarten-vd-Sande, line 148: So strange that the explain tests don't catch this.. Why is that?

seq2science/workflows/rna_seq/Snakefile
```diff
@@ -7,13 +7,13 @@ include: "../../rules/configuration.smk"
...
 include: f"{config['rule_dir']}/alignment.smk"
 include: f"{config['rule_dir']}/bam_cleaning.smk"
-include: f"{config['rule_dir']}/bigfiles.smk"
 include: f"{config['rule_dir']}/deseq2.smk"
 include: f"{config['rule_dir']}/gene_counts.smk"
 include: f"{config['rule_dir']}/get_genome.smk"
 include: f"{config['rule_dir']}/get_fastq.smk"
 include: f"{config['rule_dir']}/merge_replicates.smk"
 include: f"{config['rule_dir']}/strandedness.smk"
+include: f"{config['rule_dir']}/bigfiles.smk"
```
Maarten-vd-Sande, line 16: This isn't necessary right? At least I don't see why it would be. It's extremely minor, but nice to keep them alphabetical when possible

Maarten-vd-Sande (approved):
looks good! Not sure why the tests didn't catch the subsampe one

seq2science/workflows/rna_seq/Snakefile
Maarten-vd-Sande: :zipper_mouth_face:

seq2science/rules/configuration_explain.smk
siebrenf: it was working fine and printing "Afterwards samples were downsampled to -1 reads."

Maarten-vd-Sande: ah never mind. it's explain. i thought it was in the rule itself :)

PR #822, opened 2022-02-18
scRNA-seq to use descriptive names

siebrenf:
scRNA-seq uses `parse_samples` in the read_kb_counts script. I've changed it to use descriptive names, if given in the samples.tsv.

If this does not fit, I could also make a flag in the `parse_samples` function to return sample/replicate names. Your call @tilschaef!

**Checklist**
- [ ] I updated the CHANGELOG


siebrenf:
Great news, thank you!

Yes, it will fall back! It will try to use names from column `descriptive_names` > `technical_replicates` > `samples`.

PR #824, opened 2022-02-23
Rerun

### LINKED ISSUE #823: BUG: inconsequential CLI changes prompt reruns
**Describe the bug**
Changing the number of cores should not prompt reruns.

I'll add other options to the list if I find them.

siebrenf:
fixes #823 (partially)

**What did change**
- QC files now ignore `config["cores"]`
- CLI rerun logic will ignore params changes to trimming files.
  - it is unclear to me why there files are marked (they get marked even when removing the reference to `threads`)

**Checklist**
- [ ] I updated the CHANGELOG


seq2science/cli.py
```diff
@@ -387,12 +387,18 @@ def _run(args, base_dir, workflows_dir, config_path):
...
             sys.exit(1)
 
-        #   3. check which files would need a rerun, and clean remove files we do
-        #      not want to consider.
-        #      - genome files since provider will change to local
+        #   3. check which files would need a rerun, and exclude files we do
+        #      not want to consider:
+        #      - genome files, since provider will change to local
+        #      - trimming files, since temp() output may be gone
+        #      - samtools_presort, since temp() output may be gone
```
Maarten-vd-Sande, line 394: Why would temp files not need to be considered? In the hidden `.snakemake` folder per executed rule there is a log of the settings with which the rule was run. I'm expecting this to also work for temp files, but I can't say I've tested it... There can be important settings changed for for instance trimming..

Maarten-vd-Sande: nvm I see your comment here:

https://github.com/vanheeringen-lab/seq2science/issues/823#issuecomment-1049183719

siebrenf: excellent question... not sure 😢 

I have a s2s run with unchanged params. If I include `--skip-rerun`, no jobs are selected. Without it, all these files are marked. `config["cli_call"]` is ignored, so the `--skip-rerun` flag itself doesn't matter.

One thing that _might_ influence this, is that this run was done before these changes, so they might "remember" having `threads` in their params, which are now missing.

Maarten-vd-Sande: The thing is that if anything has to be re-run between trimming and final bam, all those steps have to be re-done since all those files are temporary files.

siebrenf:
tests failed because `OSError: [Errno 28] No space left on device`

Maarten-vd-Sande (approved):
changelog?

PR #827, opened 2022-03-08
add idr options

Maarten-vd-Sande:

small change to allow idr to be configurable

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/peak_calling.smk
+++ b/seq2science/rules/peak_calling.smk
@@ -358,7 +358,8 @@ if "biological_replicates" in samples:
             params:
                 rank=lambda wildcards: "--rank 13" if wildcards.peak_caller == "hmmratac" else "",
                 nr_reps=lambda wildcards, input: len(input),
-                reps=lambda wildcards, input: input,  # help resolve changes in input files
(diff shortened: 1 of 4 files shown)


seq2science/rules/peak_calling.smk
```diff
@@ -358,7 +358,8 @@ if "biological_replicates" in samples:
             params:
                 rank=lambda wildcards: "--rank 13" if wildcards.peak_caller == "hmmratac" else "",
                 nr_reps=lambda wildcards, input: len(input),
-                reps=lambda wildcards, input: input,  # help resolve changes in input files
+                reps=lambda wildcards, input: input,  # help resolve changes in input files,
```
siebrenf, line 361: optional: `reps=lambda wildcards, input: input, # help resolve changes in input files,` can go. `wildcards, input` already checked in the param above

siebrenf (approved):
if you tested it with a any parameter and it works, I'm good!

seq2science/rules/peak_calling.smk
Maarten-vd-Sande: I'm not sure. Since the one before only checks the length of the input, and not the actual input

PR #830, opened 2022-03-24
single cell DESeq

siebrenf:
**What's new?**
We can now use deseq2science to analyze single cell counts, if you provide groupings for all cells.

This work could be applied to the single cell workflows, but that would require some automated cluster step.

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande (approved):
LGTM! :smile: 

Only thing worth discussing here is whether it makes sense the cluster ID is in the samples.tsv or if that should be a separate file. Idk what's easier from the user point of view. And what would be easier in the future _if_ we add this to the single cell RNA-seq workflow

siebrenf:
It takes an hour or so to run, but it works! Working with 2 enormous flat text files is plain silly, but it can be done...

I'll make an issue to remind us this could be implemented, but I'm also not sure how useful that'd be.

PR #835, opened 2022-03-28
RNA-seq always produces TPM normalized counts

siebrenf:
Create some rudimentary TPMs for htseq and FeatureCounts.

**What did change**
also fixes GENCODE blacklists due to the genomepy update!

**Checklist**
- [x] I updated the CHANGELOG


seq2science/envs/pytxi.yaml
```diff
@@ -7,4 +7,4 @@ dependencies:
   - genomepy=0.11.1
   - pip
   - pip:
-    - git+https://github.com/vanheeringen-lab/pytxi.git@v0.1.1
+    - git+https://github.com/vanheeringen-lab/pytxi.git@develop
```
Maarten-vd-Sande, line 10: Maybe we put it on bioconda?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 - using awful practices to eliminate checkpoint strandedness
 - deeptools_flags renamed to deeptools_bamcoverage
 - rna-seq trackhub per base tracks by default instead of bins per 50
+- rna-seq creates a TPM table for each quantification method
```
Maarten-vd-Sande, line 32: its under the previous release :smile:

seq2science/rules/gene_counts.smk
```diff
@@ -96,7 +98,9 @@ elif config["quantifier"] == "salmon" and config["tpm2counts"] == "pytxi":
             cts=get_counts,
             fa=expand("{genome_dir}/{{assembly}}/{{assembly}}.fa", **config),
```
Maarten-vd-Sande, line 99: Is it possible to add the readme as input since it is required for this rule?

seq2science/rules/gene_counts.smk
```diff
@@ -67,6 +67,8 @@ if config["quantifier"] == "salmon" and config["tpm2counts"] == "tximeta":
             cts=get_counts,
         output:
             counts=expand("{counts_dir}/{{assembly}}-counts.tsv", **config),
+            tpms=expand("{counts_dir}/{{assembly}}-TPM.tsv",**config),
+            lengths=expand("{counts_dir}/{{assembly}}-gene_lengths.tsv",**config),
```
Maarten-vd-Sande, line 71: unrelated to the PR, but maybe there should be an input requirement for that the assembly is from ensembl? An input function should be able to do this right?

Maarten-vd-Sande: Maybe I should make an issue about it though, instead of puitting it here

seq2science/rules/gene_counts.smk
```diff
@@ -153,6 +157,21 @@ else:
                 counts.index.name = "gene"
                 counts.to_csv(output[0], sep="\t")
 
+    rule tpm_matrix:
```
Maarten-vd-Sande, line 160: :partying_face:

seq2science/rules/gene_counts.smk
```diff
@@ -153,6 +157,21 @@ else:
...
+        """
+        Create a TPM table from a counts table.
+        """
+        input:
+            cts=rules.count_matrix.output,
+            gtf=expand("{genome_dir}/{{assembly}}/{{assembly}}.annotation.gtf",**config),
+        output:
+            tpms = expand("{counts_dir}/{{assembly}}-TPM.tsv",**config),
+            lengths = expand("{counts_dir}/{{assembly}}-gene_lengths.tsv",**config),
+        log:
```
Maarten-vd-Sande, line 170: maybe add a explain message here? It's perhaps non-trivial by what/how the tpm was calculated

Maarten-vd-Sande (approved):

seq2science/envs/pytxi.yaml
siebrenf: I might when it's stable. Had to change it yesterday due to the pyproject.toml structure changes

CHANGELOG.md
siebrenf: 🙈

seq2science/envs/pytxi.yaml
Maarten-vd-Sande: Yeah but this might break future seq2science runs because develops keep changing. Maybe you can pin it on a commit instead of a branch?

siebrenf: Definetly! I'll pin it before merging.

For some reason, genomepy 0.12 is still not available on bioconda (I think there is an issue with some of the dependencies), so I can't test this PR in a clean environment yet.

seq2science/rules/gene_counts.smk
siebrenf: good point, difficult problem actually.

There are 2 ways to convert transcript IDs to gene names:
- using the GTF file (if it has gene names and transcript ID info)
- using mygene.info (if the README contains a taxonomy ID)

In reality, I don't think we want to use mygene.info. It would make it impossible to compare the counts table to the GTF, so downstream stuff (like ANANSE) becomes impossible. 
Of course, you still can do this. In this case you need the taxonomy ID, which is read from the genomepy README.txt

So now the next problem: what if somebody is using a custom genome? No readme, and you definitely dont want s2s to delete your data and try to download a new assembly (that might not even exist)!

Maarten-vd-Sande: And an input function that decides if it is downloadable from genomepy, and if yes, it gives the readme.md as input, and if not some impossible input? It's perhaps better for it to crash before starting the run than at the end?

Maarten-vd-Sande: :thinking:

Maarten-vd-Sande: Maybe Im thinking too simple about it, anyways its not part of the PR really, so an issue would also be fine?

siebrenf: Fixed. If no README.txt is available, genomepy will try to find the taxonomy online. If that fails, an error is given with a DIY solution (which should happen very rarely to begin with).

PR #842, opened 2022-04-12
Multi sce export

tilschaef:
Added SingleCellExperiment object export to Seurat and FlatFile format. The last one is especially handy for downstream processing in scanpy/scvelo. 

**What did change**
- Added scrna utility R script 
- Export sce objects to Seurat, SCE and FlatFile format.
- Added option to specify which kb velocity matrix to use for export and qc
- Simplified `read_kb_counts.R` script

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
Looks cleaner! Not much I can say about the code due to the specific libraries :) 

btw, would it be possible to take a tiny portion of some scRNA-seq dataset and make a test that actually runs?

tilschaef:
> Looks cleaner! Not much I can say about the code due to the specific libraries :)
> 
> btw, would it be possible to take a tiny portion of some scRNA-seq dataset and make a test that actually runs?

Yes, that would be a good idea given the increasing complexity of the workflow :) . We could, for example,  leave the `kb` stuff as DAG tests and then use a small droplet/plate-based UMI matrix as input for the real tests.

PR #844, opened 2022-04-12
singularity support

Maarten-vd-Sande:

Let's maybe merge this just so I can do more testing with singularity :innocent: (after a new release we get an automatic singularity container made for us), we won't add any documentation abuot singularity (yet). A nice addition is that it improves some envs of which not all requirements were properly defined.

Support for singularity should be relatively easy with this:

https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#ad-hoc-combination-of-conda-package-management-with-containers

related: https://github.com/snakemake/snakemake/pull/1588

Still needs plenty of streamlining, especially to be able to run seq2science itself from singularity, to test currently:

`export SINGULARITY_BIND="/scratch/${USER},${CONDA_PREFIX},${TMPDIR},${XDG_CACHE_HOME}"`

`seq2science run rna-seq --cores 20 --snakemakeOptions use_singularity=True conda_prefix=/scratch/sande/singularity_test/envs`

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/configuration_generic.smk
+++ b/seq2science/rules/configuration_generic.smk
@@ -1,6 +1,7 @@
 """
 all logic not related to any specific workflows should be here.
 """
+container: "docker://continuumio/miniconda3:4.4.10"
 
 import os.path
 import pickle
--- a/seq2science/rules/configuration_generic.smk
+++ b/seq2science/rules/configuration_generic.smk
@@ -1,7 +1,7 @@
 """
 all logic not related to any specific workflows should be here.
 """
-container: "docker://continuumio/miniconda3:4.4.10"
+container: "docker://quay.io/biocontainers/seq2science:0.7.2--pypyhdfd78af_0"
 
 import os.path
 import pickle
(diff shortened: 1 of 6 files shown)


seq2science/rules/configuration_generic.smk
```diff
@@ -1,6 +1,7 @@
 """
 all logic not related to any specific workflows should be here.
 """
+container: "docker://quay.io/biocontainers/seq2science:0.7.2--pypyhdfd78af_0"
```
siebrenf, line 4: this won't break anything right?

seq2science/scripts/atacshift.pl
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
```
siebrenf, line 1: might need to add/change shebangs to other scripts too

siebrenf (approved):
might have to change some `run` directives to `script`s, especially for fastq downloading. Either way, better environments is nice!

PR #852, opened 2022-04-21
enforce correct versions in s2s environment

siebrenf:
image

More errors to annoy our user with... So they won't be annoyed with weird errors later!

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,7 +22,7 @@ jobs:
           $CONDA/bin/conda install mamba=0.19.0
           $CONDA/bin/mamba env update --file requirements.yaml --name base
           $CONDA/bin/mamba install m2r2 sphinx sphinx_rtd_theme sphinx-argparse graphviz
-          $CONDA/bin/pip install .
+          $CONDA/bin/pip3 install .
 
       # -W = error on warning
       - name: Build 🔧
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ docs/content/CHANGELOG.md
 docs/content/all_rules.md
 docs/content/schemas.md
 seq2science/workflows/scrna_seq/barcodes.txt
+seq2science/envs/requirements.yaml
 
 *results/
 *benchmark/
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ docs/content/CHANGELOG.md
 docs/content/all_rules.md
 docs/content/schemas.md
 seq2science/workflows/scrna_seq/barcodes.txt
-seq2science/envs/seq2science_requirements.yaml
 
 *results/
 *benchmark/
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@ docs/content/CHANGELOG.md
 docs/content/all_rules.md
 docs/content/schemas.md
 seq2science/workflows/scrna_seq/barcodes.txt
-seq2science/envs/requirements.yaml
+seq2science/envs/seq2science_requirements.yaml
 
 *results/
 *benchmark/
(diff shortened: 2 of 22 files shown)


siebrenf:
ugh, this requires that we pack the `requirements.yaml`, which is surprisingly hard top do properly.

With the current implementation, it is placed in the base of the conda env :see_no_evil: 
Adding it to `package_data` places it in the `site-packages` folder :no_good:

Maarten-vd-Sande:
I like this! 

How did you manage it to be placed in site-packages? For me it gets ignored when t use files in root directory for package data.... How messed up would it be to put the requirements.yaml in the envs dir?

`package_data = "{'':['requirements.yaml'],'seq2science':['envs/*', 'imgs/*', 'rules/*', 'schemas/*', 'schemas/*/*', 'scripts/*', 'scripts/**/*', 'workflows/*/*']}"`

And would specifying the requirements.yaml in the MANIFEST file work?

Maarten-vd-Sande:
Maybe an idea is to take the requirements from github? Feels dirty though

Get the requirements from `f"https://github.com/vanheeringen-lab/seq2science/blob/v{seq2science.__version__}/requirements.yaml"`

siebrenf:
Changing stuff in the setup.py doesn't work:
I can get the yaml in the egg folder (`miniconda3/envs/test/lib/python3.8/site-packages/seq2science-0.7.2-py3.8.egg/`), but not to `miniconda3/envs/test/lib/python3.8/site-packages/seq2science/` (and doesn't work with `pip install .`)
```python
import os
import site

s2s_dir = os.path.join(site.getsitepackages()[0], "seq2science")
project["data_files"] = [(s2s_dir, ['requirements.yaml'])]
```


Changing stuff in the toml doesn't work either:
```toml
# did not work (yaml isn't found):
package_data = "{'':['requirements.yaml'], 'seq2science':['envs/*', 'imgs/*', 'rules/*', 'schemas/*', 'schemas/*/*', 'scripts/*', 'scripts/**/*', 'workflows/*/*']}"
# or
package_data = "{'seq2science':['requirements.yaml', 'envs/*', 'imgs/*', 'rules/*', 'schemas/*', 'schemas/*/*', 'scripts/*', 'scripts/**/*', 'workflows/*/*']}"

# installs the file in lib/python3.8/site-packages:
package_data = "{'seq2science':['../requirements.yaml', 'envs/*', 'imgs/*', 'rules/*', 'schemas/*', 'schemas/*/*', 'scripts/*', 'scripts/**/*', 'workflows/*/*']}"

# installs the file in the conda env base (miniconda3/envs/test)
data_files = "[('', ['requirements.yaml'])]"
# or 
packages = ["seq2science", ""]
package_data = "{'':['requirements.yaml'], 'seq2science':['envs/*', 'imgs/*', 'rules/*', 'schemas/*', 'schemas/*/*', 'scripts/*', 'scripts/**/*', 'workflows/*/*']}"
```

A Manifest file doesn't seem like it can help pick the target directory either.

Either we somehow automatically copy the requirements.yaml to the envs folder when making a new release, or download it from git. If you know how to do the first, that would probably be best. I'll commit a change for the latter just to try it.

Maarten-vd-Sande:
Just realized the tests won't work with the git approach :fire: 

Could we just add a mv command in the setup.py?

Maarten-vd-Sande:
A post install script might work?

https://stackoverflow.com/a/36902139/9544516

```
from setuptools.command.install import install

class PostInstallCommand(install):
    """Post-installation for installation mode."""
    def run(self):
        install.run(self)
        # PUT YOUR POST-INSTALL SCRIPT HERE or CALL A FUNCTION
```

PR #856, opened 2022-04-28
move trackhub to script (improves logging)

siebrenf:
logging in a `run` directive is busted. This seems to work...

**Checklist**


Maarten-vd-Sande (approved):
Cool. Snakemake needs to fix that `run` directives get run separately. I'm assuming this works :smile:

PR #858, opened 2022-04-29
global variables

siebrenf:
Keeping track of global variables is kinda hard. To make it more clear when you're dealing with a global or a local, I made most globals upper case, following PEP8.

There are some exceptions, mainly `samples`, `treps` and `breps` and snakemake's own globals (`config`, `rules` and `workflow`). We could make the samples-related stuff upper case too, just let me know :) 

A bunch code in configuration_generic() has been moved into functions (that often use globals). These could be moved to the utils(?)

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/deseq2.smk
+++ b/seq2science/rules/deseq2.smk
@@ -1,8 +1,7 @@
 """
 all rules/logic
 """
-
-import sys
+import os
 
 
 """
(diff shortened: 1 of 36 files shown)


Maarten-vd-Sande (approved):
Nice, some necessary cleanup :)

No one ever reads this, but could you update the variables here as well?
https://github.com/vanheeringen-lab/seq2science/blob/master/docs/CONTRIBUTING.md

seq2science/rules/deseq2.smk
```diff
@@ -47,7 +46,7 @@ rule deseq2:
     log:
         expand("{log_dir}/deseq2/{{assembly}}-{{contrast}}.diffexp.log", **config),
     message:
-        explain_rule("deseq2")
+        EXPLAIN.get("deseq2", "")
```
Maarten-vd-Sande, line 49: wouldn't it have been easier to make this a defaultdict?

siebrenf: it is now :)

PR #867, opened 2022-05-23
allow samples.tsvs with multiple assemblies

siebrenf:
When using `deseq2science`, you can now optionally specify which assembly is being used.
By default the first assembly is expected (as it was before).

**Checklist**
- [x] I updated the CHANGELOG


Maarten-vd-Sande (approved):

Maarten-vd-Sande:
Anansnake doesn't work with newer seq2science versions unfortunately since the `parse_samples` has been moved. I'll check if I can put it back in util?

siebrenf:
it wasn't moved but renamed. I made a PR that we can use when these changes are released :)

PR #873, opened 2022-05-31
index shouldnt be temp

Maarten-vd-Sande:

indexes shouldn't be temp

**Checklist**
- [x] I updated the CHANGELOG


siebrenf:
Ideally they should be temp, unless they output to the `final_bam_fir`, or are `.cram.brai`.
If that's not possible, I'm not sure its worth keeping all the intermediate indexes just to keep the final ones, since they're so easy to make...

PR #877, opened 2022-06-07
samples in the cloud fix

Maarten-vd-Sande:

Some samples crash because they've been moved to the :cloud: , newest pysradb has no issues with that!

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
That looks relatively painless!

PR #878, opened 2022-06-07
Update requirements.yaml

siebrenf:
Conda 4.13 is now out on conda-forge. Perhaps that solves the issues we had last week...

### DIFF
--- a/requirements.yaml
+++ b/requirements.yaml
@@ -6,8 +6,8 @@ channels:
 dependencies:
   - conda-forge::python=3.8
   - bioconda::snakemake-minimal=7.7.0
-  - conda-forge::conda=4.11.0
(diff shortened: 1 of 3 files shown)


Maarten-vd-Sande (approved):
I think this was the actual problem: https://github.com/vanheeringen-lab/seq2science/commit/c33cf8425156284392cc560b688387d739318e33

But newer versions are "always" good :smile:

siebrenf:
btw, not sure if this is interesting for us, but they... integrated mamba with conda>=4.12. It's experimental right now, and not on by default yet.

https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community

Maarten-vd-Sande:
:partying_face:

PR #880, opened 2022-06-08
add niceness

Maarten-vd-Sande:

Configurable niceness setting

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 ### Fixed
 
 - samples moved to the cloud on SRA can be downloaded again with newest pysradb.
+- issue with generating a trackhub index with tiny transcripts (we just remove them :) )
+- bug with sralite files having the wrong file extension so they're not recognized as downloaded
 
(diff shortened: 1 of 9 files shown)


siebrenf (commented):
Seems nice in theory, but I have a lot of questions about how this works...
- This applies only to shell rules right? Does that mean shell rules would ordinarily have no nicess when running `nice seq2science`?
- What happens if you use `nice seq2science` and this parameter?
- Should there be a default niceness in de schema? 
- Or should it detect if you are running `nice seq2science`?

Maarten-vd-Sande:
- yes, only works for shell rules. But if you run `nice seq2science` it applies to everything. Problem is that nice seq2science + slurm don't work together, so everything gets submitted without niceness
- Haven't tried it, but im guessing the niceness in the config overwrites the nice you start the command with
- I was thinking about this, but I don't think there should be a default. As that would be unexpected behaviour if you run `nice seq2science` yourself.
- I was looking into this, I'm not sure if that's possible. Probably possible but not very easy

siebrenf (approved):
OK, without a default niceness there shouldn't be any issues even if this parameter and `nice` don't work together :)

Issue #886, opened 2022-06-20
BUG: softmask_track_1 gets stuck


siebrenf:
**Describe the bug**
Jenkins sometimes gets stuck on rule softmask_track_1 [rule](https://github.com/vanheeringen-lab/seq2science/blob/5c13d49417ccd76cde90aace2279effdf4928726/seq2science/rules/trackhub.smk#L128). 

http://ocimum.science.ru.nl/jenkins/blue/organizations/jenkins/seq2science/detail/PR-884/1/pipeline/
http://ocimum.science.ru.nl/jenkins/blue/organizations/jenkins/seq2science/detail/develop/250/pipeline/
http://ocimum.science.ru.nl/jenkins/blue/organizations/jenkins/seq2science/detail/develop/148/pipeline/97

**possible solutions?**
- end the rule with `pool.join()`
- add a timeout + retry
- switch from imap to async

Maarten-vd-Sande:
I think this was solved right?

PR #896, opened 2022-09-22
Update trackhub.smk

siebrenf:
Suggested fix for #886 by Sybren

**What did change**
added a timeout to `mp.Pool.imap_unordered()`. Not sure if this timeout is per iteration or for the whole `imap` (I assumed the former).

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ### Fixed
 
-- downloading fastq from ena directly fixed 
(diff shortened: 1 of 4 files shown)


Maarten-vd-Sande (approved):
:+1: 

Isn't 10 mins a bit much though? Better than no timeout I guess :smile:

siebrenf:
Yeah, it's an arbitrary length. That should allow even the most ridiculously long contigs to get mapped (or all contigs, in case the timeout is pool wide, not function wide). Either way, Jenkins will be happier for it ;)

PR #913, opened 2022-11-16
pretty docs DAGs

siebrenf:
RNA-seq:
image

ATAC-seq:
image

scATAC-seq:
image


docs/scripts/clean_dags.py
```diff
@@ -0,0 +1,299 @@
+import re
+import sys
+import networkx as nx
+
+rules_to_hide = [
```
Maarten-vd-Sande, line 5: would it make sense to change this into a rules to keep? :innocent:

Maarten-vd-Sande (approved):

Maarten-vd-Sande:
Should we mention in the docs that this is a "simplified" graph?

docs/scripts/clean_dags.py
siebrenf: yea it would :cry:

Maarten-vd-Sande: It's also easier if rules get added later down the line :grimacing:

PR #918, opened 2022-11-21
update snakemake

Maarten-vd-Sande:

update snakemake to newest version. Seems to work for me. Just one weird thing is that sometimes stuff gets printed and not logged and I don't have a clue why. It's kinda non-informative, so doesn't matter if it doesn't end up in logfile anyways.. But still...

Also I enabled the debug logging when --debug is on. Could be handy for ourselves in case we want to print stuff during development but not for users..?
image

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/cli.py
+++ b/seq2science/cli.py
@@ -339,6 +339,7 @@ def _run(args, base_dir, workflows_dir, config_path):
         "unlock": args.unlock,
         "cleanup_metadata": args.cleanup_metadata,
         "force_incomplete": args.rerun_incomplete,
+        "rerun_triggers": ["mtime", "input", "software-env"],
     }
 
     # get the additional snakemake options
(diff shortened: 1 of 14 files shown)


seq2science/cli.py
```diff
@@ -387,13 +388,13 @@ def _run(args, base_dir, workflows_dir, config_path):
...
+    setup_seq2science_logger(parsed_args, args.debug)
     log_welcome(logger, args.workflow)
 
     if args.debug:
         # dump the parsed args as readable json
         import json
-        logger.info(json.dumps(parsed_args, sort_keys=True, indent=2))
+        logger.debug(json.dumps(parsed_args, sort_keys=True, indent=2))
 
     if not args.skip_rerun or args.unlock or args.cleanup_metadata is not None:
```
siebrenf, line 399: What do lines 398 to 444 still add with the newest snakemake?

Maarten-vd-Sande: Yes I thought they could be removed as well at first. But we still somehow need to catch the `regex_patterns` since the provider changes for the assemblies.... :upside_down_face: 

Unless we somehow remove the provider param?

PR #928, opened 2022-11-28
Bugfix bioreps

Maarten-vd-Sande:

fix bug is combine_bioreps

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 ### Fixed
 
 - clear error when specifying unavailable trimmer (#888)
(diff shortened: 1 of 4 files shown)


siebrenf (approved):
- nice idea to skip the calculations without biological replicates
- your solution still assumed "descriptive_name" in the samples.tsv.
- in the rule, you pipe the processed samples dataframe from snakemake directly to the script, so you don't have to process it again inside the script. (with `dense_samples()`)
- I changes some lines to make them more readable (in the script and the configurations)

siebrenf:
You can merge if you approve of my changes ;)

Maarten-vd-Sande:
Nice! I don't think this gets tested in the auto-tests. Did you confirm this works?

PR #929, opened 2022-11-28
fix schema for deseq in multiqc

Maarten-vd-Sande:

Ordering wasn't correct

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 ### Fixed
 
 - clear error when specifying unavailable trimmer (#888)
+- fixed the ordering of deseq2 output in the multiqc report
 
 ## [0.9.6] - 2022-10-31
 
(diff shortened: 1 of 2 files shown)


siebrenf (commented):
from the rules:
`expand("{qc_dir}/deseq2/{{assembly}}-{{contrast}}.pca_plot_mqc.png", **config),`
`expand("{qc_dir}/deseq2/{{assembly}}-{{contrast}}.combined_ma_volcano_mqc.png", **config),`

what was wrong, and how does this fix it? 
There is a period between the contrast and the suffix that isn't changes by the `replace`, so I think this shouldn't work?

Maarten-vd-Sande:
Ok while I was answering your question about that multiqc internally changes dots into underscores I noticed indeed it doesn't work fully. This makes it so that the ordering works, BUT the description is gone :thinking: 

Thanks! I'll make it so that the description is also still there :sweat_smile:

Maarten-vd-Sande:
This seems not necessary?!

PR #931, opened 2022-11-30
slightly cleaner genomepy error

Maarten-vd-Sande:

I dislike the genomepy errors printed as stacktrace and its unclear to the user what is going on! This should help :smile: 

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
I'm struggling with the same issue sometimes: massive stacktraces when only human readable messages would be more informative. 

Your solution will still give a stacktrace on `sys.exit(1)` however. You could use `os._exit(1)  # noqa` to get a stacktrace-free exit. (If you know a better way, let me know ;) )

Maarten-vd-Sande:
Good catch, you're on fire :fire: ! I keep forgetting this still gives a stacktrace :facepalm:

PR #935, opened 2022-12-01
optional motif2factors

Maarten-vd-Sande:
**What problem is the PR solving / What's new?**

Flag whether motif2factors should be run. I was working with some fly data and I didn't want it :smile: 

**Checklist**
- [x] I made a PR to develop (not master)
- [ ] If applicable: I updated the docs
- [x] I updated the CHANGELOG
- [ ] These changes are covered by the tests

PR #936, opened 2022-12-01
explain tests failing

Maarten-vd-Sande:

Explain didn't work because of a wrong gimmemotifs lookup (solved in #935). However this did not get caught by the auto tests. This should catch it.

**Checklist**
- [ ] I updated the CHANGELOG


Maarten-vd-Sande:
Not working explain tests are now caught as well with an empty return. Cause for empty return fixed in #935 
image

siebrenf (approved):

PR #948, opened 2023-01-11
Update configuration_generic.smk

Maarten-vd-Sande:

When downloading samples but there is an assembly column in the samples file, errors happen :sob: 

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


siebrenf (changes requested):
The changelog message and the code change to not match?

In this PR, the condition is still True if `"assembly" in samples`. Did you forget to remove that part? Just `if WORKFLOW != "download_fastq":` should work...

Maarten-vd-Sande:
Good point :sweat_smile: ... Shouldve been an `and`. But I'll just remove the second half since they are mutually exclusive anyways

siebrenf (approved):

PR #950, opened 2023-01-18
Update requirements.yaml

Maarten-vd-Sande:

It works with mambaforge :partying_face: ! Thanks @siebrenf 

Somehow doesnt work in the base env anymore so we can't split over multiple "checkpoints", but idc

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           python-version: 3.8
 
-      - name: Install dependencies 🔨
+      - name: Install dependencies 🔨 & Build 🔧
         run: |
           # setup mambaforge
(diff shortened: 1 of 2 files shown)


siebrenf:
I don't know if this is a temporary issue, or if conda and mamba are drifting apart now that the libmamba-solver exists, but I couldn't install mamba either for the genomepy docs. [My solution there](https://github.com/vanheeringen-lab/genomepy/blob/master/.github/workflows/docs.yml) was to install micromamba, which was pretty fast and easy.

Maarten-vd-Sande:
Ok I'll try that, thanks :smile:

siebrenf (approved):
the name (and emoji) of the step could be updated, but only if you feel like it :)

PR #952, opened 2023-01-24
Gsa support

Maarten-vd-Sande:

Support for the Chinese GSA (genome sequence archive). We'll only support CRX numbers for them :smile: 

**Checklist**
- [x] I updated the CHANGELOG


seq2science/rules/get_fastq.smk
```diff
@@ -63,11 +63,12 @@ rule run2sra:
 
 
 # ENA > SRA
-ruleorder: ena2fastq_SE > sra2fastq_SE
-ruleorder: ena2fastq_PE > sra2fastq_PE
+ruleorder: ena2fastq_SE > gsa2fastq_SE > sra2fastq_SE
+ruleorder: ena2fastq_PE > gsa2fastq_PE > sra2fastq_PE
```
siebrenf, line 67: i assume gsa is used least, so maybe put it last?

siebrenf (approved):

seq2science/rules/get_fastq.smk
Maarten-vd-Sande: doesn't really matter, they are mutually exclusive

PR #954, opened 2023-01-30
explanation fixed

Maarten-vd-Sande:

I noticed a lot of empty explanations.... This should fix that :smile: 

image

I also improved the tests

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
weird but ok

Maarten-vd-Sande:
Yeah idk why it was working before, but the problem was that it couldn't lock the directories so it crashed. This wasnt caught with the explain tests as it wasn't running in locked directories. But when running a real run and if then the explain rule happenend it would complain it can't lock...

No clue why it changed, but now it works again and is properly tested

PR #955, opened 2023-02-03
localrules

Maarten-vd-Sande:

better cluster support

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
not sure what this does but OK

Maarten-vd-Sande:
localrules are not run on a node when using e.g. slurm. Since downloading does not use compute I think it makes more sense to not use nodes.

PR #965, opened 2023-03-21
doesn't work as grouped jobs are ignored when not submitting to cluster

Maarten-vd-Sande:

@simonvh I was thinking something like this. Does NOT work apparently because grouped jobs only work in cluster mode. Perhaps it's possible that gets turned on though through some settings. I do not want this as a default, so it would still depend on a config flag.

**Checklist**
- [ ] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/get_fastq.smk
+++ b/seq2science/rules/get_fastq.smk
@@ -1,11 +1,14 @@
 """
 all rules/logic related downloading public fastqs should be here.
 """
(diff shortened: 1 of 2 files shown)


simonvh:
Yeah, I couldn't find a setting to enable grouping of rules locally. I think this will also result in files being copied from the shadow dir back to the output dir, once the rule is finished, no? This means a lot of copying. I'm not sure that will improve performance either.

PR #966, opened 2023-03-21
clear error, fixes #962

### LINKED ISSUE #962: BUG: Incorrect SRA files (RNA-seq)
**Describe the bug**

Some SRA files are incorrect (either due to submitter not supplying correct files, or SRA not having correct annotation). For example, this file:
[SRR18972600](https://trace.ncbi.nlm.nih.gov/Traces/?view=run_browser&acc=SRR18972600&display=metadata) (GSM6077643). This marked as PE, but `fastq-dump` only gets one fastq file.

**To Reproduce**

RNA-seq workflow, with the following `samples.tsv`:

```
sample  assembly
GSM6077643     GRCh38
```

**Expected behavior**
I don't know! Either an informative error message (i.e. two fastq files should be there, there's only one, this can be checked) or the file being treated as single-end. The second option may not be correct, and I assume this may also be hard to implement. I would guess an error message would already be good, because the error is not easily identifiable from the log, but only when you manually repeat the steps.

One thing I noticed that these are not sequenced with Illumina, but with DNBSEQ-G400. 

Low priority bug I would say, but something to be aware of.

Maarten-vd-Sande:

Clearer error message when single-end is annotated as paired-end. Fixes #962 

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/seq2science/rules/get_fastq.smk
+++ b/seq2science/rules/get_fastq.smk
@@ -150,10 +150,14 @@ rule sra2fastq_PE:
         --threads {threads} --split-e --skip-technical --dumpbase \
         --readids --clip --read-filter pass --defline-seq '@$ac.$si.$sg/$ri' \
         --defline-qual '+' --gzip >> {log} 2>&1
+        
+        # check if the files exist
+        if ! compgen -G "{output.tmpdir}/*_1*" > /dev/null ; then echo "ERROR: Couldn't find read 1.fastq after dumping! Perhaps this is not a paired-end file?\n" >> {log} 2>&1; fi
+        if ! compgen -G "{output.tmpdir}/*_2*" > /dev/null ; then echo "ERROR: Couldn't find read 2.fastq after dumping! Perhaps this is not a paired-end file?\n" >> {log} 2>&1; fi
 
         # rename file and move to output dir
-        mv {output.tmpdir}/*_1* {output.fastq[0]}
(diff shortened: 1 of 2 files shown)


siebrenf (approved):

seq2science/rules/get_fastq.smk
```diff
@@ -150,10 +150,14 @@ rule sra2fastq_PE:
         --threads {threads} --split-e --skip-technical --dumpbase \
         --readids --clip --read-filter pass --defline-seq '@$ac.$si.$sg/$ri' \
         --defline-qual '+' --gzip >> {log} 2>&1
+        
+        # check if the files exist
+        if ! compgen -G "{output.tmpdir}/*_1*" > /dev/null ; then printf "ERROR: Couldn't find read 1.fastq after dumping! Perhaps this is not a paired-end file?\n" >> {log} 2>&1; fi
+        if ! compgen -G "{output.tmpdir}/*_2*" > /dev/null ; then printf "ERROR: Couldn't find read 2.fastq after dumping! Perhaps this is not a paired-end file?\n" >> {log} 2>&1; fi
```
Maarten-vd-Sande, line 156: I did not know that echo doesn't work :sweat_smile: ... I did just now realize that the newline isn't even necessary, as echo automatically makes a newline. Do we know if printf is standard in bash?

PR #970, opened 2023-03-22
update prefetch

siebrenf:
Nope, you really can't pipe `prefetch` output. But we can update the tool! Prefetch received bugfixes and can now accept and unlimited max-size. And we can clean up the yaml a bit (pigz isn't used anymore?)

### DIFF
--- a/seq2science/rules/get_fastq.smk
+++ b/seq2science/rules/get_fastq.smk
@@ -33,6 +33,7 @@ rule run2sra:
         # and remove the top level folder since prefetch will assume we are done otherwise
         mkdir -p {params.outdir}; cd {params.outdir}; rm -r {wildcards.run}
 
+        # TODO: for loop can be removed if we dont see the debug message
         # three attempts
         for i in {{1..3}}
         do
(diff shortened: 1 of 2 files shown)


seq2science/rules/get_fastq.smk
```diff
@@ -33,6 +33,7 @@ rule run2sra:
         # and remove the top level folder since prefetch will assume we are done otherwise
         mkdir -p {params.outdir}; cd {params.outdir}; rm -r {wildcards.run}
 
+        # TODO: for loop can be removed if we dont see the debug message
```
Maarten-vd-Sande, line 36: I don't understand?

Maarten-vd-Sande: I actually don't understand what the original code does actually haha

Im not sure actually why this lock is here. I guess some filelock issues with many parallel starts. But why only this rule specifically? :thinking:

seq2science/rules/get_fastq.smk
```diff
@@ -43,21 +44,26 @@ rule run2sra:
...
             mv {params.outdir}/{wildcards.run}.sra {output}
         fi
 
+        # TODO: section can be removed if we dont see the debug message
         # If an sralite file was downloaded instead of a sra file, just rename it
         if [[ -f "{params.outdir}/{wildcards.run}.sralite" ]]; then
+            echo "DEBUG: renaming SRAlite" >> {log}
             mkdir -p {params.outdir}/{wildcards.run}
             mv {params.outdir}/{wildcards.run}.sralite {output}
         fi
```
Maarten-vd-Sande, line 69: I wouldn't be so sure that just because we don't have these issues with our test data, that we don't get them in general. The sralite one used to be specific to some accessions if I remember correctly

seq2science/rules/get_fastq.smk
```diff
@@ -301,12 +307,10 @@ rule runs2sample:
...
         get_runs_from_sample,
     output:
-             expand("{fastq_dir}/{{sample}}{{suffix}}", **config) if keep_fastqs else \
-        temp(expand("{fastq_dir}/{{sample}}{{suffix}}", **config))
+        expand("{fastq_dir}/{{sample}}{{suffix}}", **config) if keep_fastqs else \
+            temp(expand("{fastq_dir}/{{sample}}{{suffix}}", **config))
     log:
         expand("{log_dir}/run2sample/{{sample}}{{suffix}}.log", **config),
-    benchmark:
-        expand("{benchmark_dir}/run2sample/{{sample}}{{suffix}}.benchmark.txt", **config)[0]
```
Maarten-vd-Sande, line 309: Don't see a reason why not to keep this?

seq2science/rules/get_fastq.smk
siebrenf: the top issue should be fixed in prefetch now. I left it in because I agree ;)

Maarten-vd-Sande (approved):

PR #972, opened 2023-03-28
faster envs

siebrenf:
speed up conda by:
- using the latest compatible versions of conda & mamba
- dropping the usage of the `defaults` conda repo
- using the latest snakemake

Some other packages have been updated as well, mostly to bypass Conda channel priority errors. Exceptions: tabulate (supports newer python versions), genomepy (same version everywhere now) and matplotlib (using matplotlib-base now)

Also included: a rough fix for CRX request rate limit (`time.sleep(0.1)`), which is not 100% effective, but is a big improvement.

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/.github/workflows/continuousdeployment.yml
+++ b/.github/workflows/continuousdeployment.yml
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
-    - uses: actions/setup-python@v1
+    - uses: actions/setup-python@v4
       with:
         python-version: '3.10'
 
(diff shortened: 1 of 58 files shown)


Maarten-vd-Sande:
Are we confident that this is "enough" testing? I'm not sure if we are skipping some error-prone rules or not in our tests :thinking:

Maarten-vd-Sande:
Not a real benchmark, but on master it takes 3:47 minutes to generate all environments, and on this branch 2:36. So that's pretty significant, nice!

Maarten-vd-Sande (approved):

siebrenf:
ATAC-seq and RNA-seq look good (both tested). The single-cell workflows have not really been changed (I pinned a python version somewhere, and the env can be generated).

The bugfixes I added should not have anything to do with the env changes! It seem code just breaks the moment you look away...

PR #981, opened 2023-04-25
small text

Maarten-vd-Sande:
Twice effective genome size text...


siebrenf (approved):
probably from back when checkpoints were fixed, and then broke again.

PR #987, opened 2023-05-03
workflow explanation by default outputted

Maarten-vd-Sande:

End of workflow runs a explain. Do we like this?

image

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (commented):
The creepy dancing guy is nice, because it attracts attention and his speech bubble directs you to the right places.

But now there's an additional wall of text, that can already be found in the MultiQC (which the creepy dancer directs you too), and on the CLI. So I think this might be a bit much?

PR #989, opened 2023-05-17
gimme maelstrom without XDG_CACHE_DIR set

Maarten-vd-Sande:

gimme maelstrom needs dependency that was removed with latest update and if cache not set it didnt work the previous implementation 👼 

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
xgboost is an optional dependency, but it makes sense to use it in s2s!

no idea what the XDG cache stuff does, but if it works I'm happy

PR #990, opened 2023-05-19
Update subread.yaml

Maarten-vd-Sande:

Single-ended reads need different call that paired-end to featurecounts. I can't be arsed to change it so this is easiest :innocent: 

**Checklist**
- [ ] I updated the CHANGELOG


siebrenf (approved):
fair

Issue #995, opened 2023-06-02
BUG: latest numpy version and IDR inconsistency


apposada:
Hi,

First time trying this pipeline and it works very well! I had one minor setback and I was not sure under which category I should post it (I think Bug report is the best-suited?)

I was running seq2science atac-seq with two samples of two biological replicates each, using IDR to merge them, and I got this error in the log of IDR:

```
/mnt/sda/alberto/programs/miniconda3/envs/seq2science_venv/lib/python3.10/site-packages/seq2science/.snakemake/c90ff8cbd1e27eafcd9b38725c49c49c_/bin/idr --samples /mnt/sda/alberto/projects/smed_cisreg/outputs/bulk_atac/gehrke_2019/results/macs2/Smed-sample1_peaks.narrowPeak /mnt/sda/alberto/projects/smed_cisreg/outputs/bulk_atac/gehrke_2019/results/macs2/Smed-sample2_peaks.narrowPeak --output-file /mnt/sda/alberto/projects/smed_cisreg/outputs/bulk_atac/gehrke_2019/results/macs2/Smed-tail_0h_peaks.tmp.narrowPeak
Traceback (most recent call last):
  File "/mnt/sda/alberto/programs/miniconda3/envs/seq2science_venv/lib/python3.10/site-packages/seq2science/.snakemake/c90ff8cbd1e27eafcd9b38725c49c49c_/bin/idr", line 10, in <module>
    idr.idr.main()
  File "/mnt/sda/alberto/programs/miniconda3/envs/seq2science_venv/lib/python3.10/site-packages/seq2science/.snakemake/c90ff8cbd1e27eafcd9b38725c49c49c_/lib/python3.10/site-packages/idr/idr.py", line 846, in main
    r1, r2 = build_rank_vectors(merged_peaks)
  File "/mnt/sda/alberto/programs/miniconda3/envs/seq2science_venv/lib/python3.10/site-packages/seq2science/.snakemake/c90ff8cbd1e27eafcd9b38725c49c49c_/lib/python3.10/site-packages/idr/idr.py", line 299, in build_rank_vectors
    return ( numpy.array(rank1, dtype=numpy.int),
  File "/mnt/sda/alberto/programs/miniconda3/envs/seq2science_venv/lib/python3.10/site-packages/seq2science/.snakemake/c90ff8cbd1e27eafcd9b38725c49c49c_/lib/python3.10/site-packages/numpy/__init__.py", line 305, in __getattr__
[...]

PR #996, opened 2023-06-04
idr requires numpy<1.20

Maarten-vd-Sande:

see #995 

**Checklist**
- [x] I updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All changed fall under either one of these types: `Added`, `Changed`, `Deprecate
 
 ## [Unreleased]
 
+### Fixed
+
+- idr bug with numpy dependency > 1.20
+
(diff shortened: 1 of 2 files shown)


siebrenf (approved):
Do we still need this with your awesome bioconda fix?

Alternatively, we could also pin the new idr build:
```
- bioconda::idr=2.0.4.2=py39hec16e2b_10
```

(N)either way is good :)

Maarten-vd-Sande:
Yeah I wasn't sure what would be best or necessary. For the new build I pinned the python version lower than 3.9, I am not sure what conda prefers. The newer build or the same python (3.10) version. So I thought just to be safe I pin numpy in the idr environment as well. Pinning the idr should also work. I think the downstream effect is (hopefully) all the same though. :smile:

PR #999, opened 2023-07-14
Update util.py

Maarten-vd-Sande:

>Changed in version 3.5: The etype argument is ignored and inferred from the type of value.

Crashes now. So I guess removing it solves the problem...??

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (approved):
lgtm

PR #1008, opened 2023-09-11
fix issue when colors in table

Maarten-vd-Sande:

Colors get spaces in them as a tuple (12, 42, 32) so the qc report table gets fked :smile: . By specifying that a column has AT LEAST two spaces we circument this problem.

**Checklist**
- [x] I updated the CHANGELOG


siebrenf (changes requested):
The colors column also accepts colors by name. And I'm not sure this holds for the regular columns? I'll have a look.

